PR #27, opened 2016-08-08
ZONE module

bvandersloot:
Performs ALOOKUP on all domains in a root zone file. This attempts to skip the first hop where possible, not hitting local or root DNS servers. It instead uses glue records and NS information in the root zone file to connect directly to the nameserver.

This is not general, and currently only works with ALOOKUP.

### DIFF
--- a/conf.go
+++ b/conf.go
@@ -32,6 +32,8 @@ type GlobalConf struct {
 	MetadataFilePath string
 
 	NamePrefix string
+
+	Module string
 }
 
 type Metadata struct {
--- a/conf.go
+++ b/conf.go
@@ -54,8 +54,8 @@ type Result struct {
 }
 
 type TargetedDomain struct {
-	Domain     string `json:"domain"`
-	Nameserver string `json:"nameserver"`
+	Domain      string   `json:"domain"`
+	Nameservers []string `json:"nameservers"`
 }
 
 type Status string
--- a/conf.go
+++ b/conf.go
@@ -53,6 +53,11 @@ type Result struct {
 	Data        interface{} `json:"data,omitempty"`
 }
 
+type TargetedDomain struct {
+	Domain     string `json:"domain"`
+	Nameserver string `json:"nameserver"`
+}
+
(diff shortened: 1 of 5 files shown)


lookup.go
```diff
@@ -73,17 +73,35 @@ func doLookup(g *GlobalLookupFactory, gc *GlobalConf, input <-chan string, outpu
...
+			prefixedName, changed := makeName(rawName, gc.NamePrefix)
+			if changed {
+				res.AlteredName = prefixedName
+				targeted.Domain = prefixedName
+			}
+			query, err := json.Marshal(targeted)
+			lookupName = string(query)
+			if err != nil {
+				log.Fatal("Internal Error: %s", err)
+			}
```
zakird, line 90: Why is this in the framework and not in the module itself?

conf.go
```diff
@@ -51,6 +53,11 @@ type Result struct {
 	Data        interface{} `json:"data,omitempty"`
 }
 
+type TargetedDomain struct {
+	Domain      string   `json:"domain"`
+	Nameservers []string `json:"nameservers"`
+}
+
```
zakird, line 60: Does this need to be defined globally?

lookup.go
```diff
@@ -140,12 +158,56 @@ func doInput(in chan<- string, path string, wg *sync.WaitGroup) error {
...
+			in <- string(query)
+		}
+	} else {
+		s := bufio.NewScanner(f)
+		for s.Scan() {
+			in <- s.Text()
+		}
+		if err := s.Err(); err != nil {
+			log.Fatal("input unable to read file", err)
+		}
```
zakird, line 210: Same question as earlier---why is this in the framework and not in the zone module itself?

bvandersloot:
Looking to make changes to make the zonefile input option more generic.

zakird:
@bvandersloot I'm going to close this given our conversation since I don't want someone to spend time to review given the rearrange. Let's just open a new PR when you're ready?

PR #36, opened 2016-08-13
ZONE module

bvandersloot:
This allows the zonefiles to be used as inputs, passing a *dns.Token to the callback as an alternative, and generalizes all other modules to obey the new interface. A new module, ZONE, is added that takes the zonefile, and does ALOOKUPs for each domain to its specified nameservers, making use of glue records. If IPs are found, more requests are not made to the subsequent nameservers.

The only catch is that we can get >1 output per domain, e.g. when the first nameserver fails. The only way to get around this would be to have some way to know that no more names are coming, and do more chan magic.

Looking for feedback.


zakird:
> The only catch is that we can get >1 output per domain, e.g. when the first nameserver fails. The only way to get around this would be to have some way to know that no more names are coming, and do more chan magic.

Why can't you just not output the first one? Couldn't we just lock on that name? 

Otherwise, couldn't we just pass a list of tokens into the module at once that are all for the same domain and let the DoLookup function take care of that logic internally?

bvandersloot:
> Why can't you just not output the first one? 

Because there could be N nameservers for one domain

> Couldn't we just lock on that name?

How could a token know it is the last one in this scheme?

> Otherwise, couldn't we just pass a list of tokens into the module at once that are all for the same domain and let the DoLookup function take care of that logic internally?

We had this before, but you did not like how it led the framework to be less general- see the last PR for ZONE. I'm not particularly sold either way right now.

zakird:
Yeah, I see what you mean. How ugly.

zakird:
Re framework being less general, my concern was over you have static if statements in the framework that were dependent on a module, not how much data was passed to the DoLookup (or similar) call at any given time.

But, I do agree it's a bit ugly for the framework to group together. I think we can definitively say that sometimes outputting bad or repeated data is not a good option.

zakird:
Maybe I'm missing something, but it seems like this could easily be avoided by just checking if it has been output already before outputting it:

```
def DoLookup(token)
    if domain in hashcache:
        return
    retv = lookup(domain)
    if not retv.success:
        return
    output_already = cachehash.chkadd(domain, retv)
    if output_already:
        return
    return retv
```

bvandersloot:
What if the first output is a failure?

This may work if we forward the result instead of a bool though...

zakird:
If the first attempts fails, then it wouldn't get added to the cachehash and the second one would output it?

bvandersloot:
What about where two requests happen in parallel, and the one that returns first is a failure?

zakird:
I'm pretty sure that situation is handled here. You don't add anything to cachehash if you're not successful. I have updated the example above to be more explicit.

bvandersloot:
I do not see a path where a failure is outputted, in this more explicit edit. It will always fail out at the success check.

zakird:
This is true---failures are not currently output. Do we need to output
failures though? This wouldn't be horrible either. We'd just need to have
an extra state in the cache hash and output failure on eviction.

bvandersloot:
That could be it. Not outputting is easy to change on this PR. I am just not sure about how to add a general CB that can output in golang. I guess I can give it a shot.

zakird:
What do you mean general CB? like in the cachehash library? You could either add a manual evict call or have the library take a function pointer on initialization.

bvandersloot:
Yeah, and one that will actually be useful. I haven't put much thought into it, I just see no register CB in golang cachehash.

zakird:
we'll have to add it

modules/zone/zone.go
```diff
@@ -198,11 +201,11 @@ type GlobalLookupFactory struct {
 func (s *GlobalLookupFactory) AddFlags(f *flag.FlagSet) {
 	f.BoolVar(&s.Subfactory.IPv4Lookup, "ipv4-lookup", true, "perform A lookups for each server")
 	f.BoolVar(&s.Subfactory.IPv6Lookup, "ipv6-lookup", true, "perform AAAA record lookups for each server")
-	f.IntVar(&s.CacheSize, "cache-size", 100000, "number of records to store in cache of successes, preventing duplicate lookups")
 }
 
 func (s *GlobalLookupFactory) Initialize(c *zdns.GlobalConf) error {
 	s.GlobalConf = c
+	s.CacheSize = c.Threads * 100
```
zakird: 100 seems high to me. Can you check--- are there any domains with more than 10 name servers? I suspect that 25 would be more reasonable. Maybe we should also allow this as a CLI flag.

PR #66, opened 2016-08-26
Fix file designation text in error msgs

paul-pearce:


zakird:
Thanks for catching this. Looks good to me. 👍

PR #71, opened 2016-09-08
Fix -log-file so that it can log to a file.

paul-pearce:
Currently the file is opened via os.Open() which is a read-only shortcut to os.OpenFile(). This generates error if a file doesn't exist and bad stuff if you actually try to log something.


zakird:
👍

PR #74, opened 2016-09-20
Zone processing fix

bvandersloot:


modules/zone/zone.go
```diff
@@ -37,42 +36,6 @@ type Lookup struct {
...
-	res.Name = domain
-	res.Status = string(zdns.STATUS_ERROR)
-	jsonRes, err := json.Marshal(res)
-	if err != nil {
-		log.Fatal("Unable to marshal JSON result", err)
-	}
-	c.OutputFile.WriteString(string(jsonRes) + "\n")
-}
-
 func LookupHelper(domain string, nsIPs []string, lookup *alookup.Lookup) (interface{}, zdns.Status, error) {
```
zakird, line 39: Does this need to be exported?

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
-		s.Factory.Factory.CHmu.Unlock()
-		for unsolved := range notify {
-			if unsolved {
-				proceed = true
-				break
-			} else {
-				notify <- false
-				break
-			}
+	for {
```
zakird, line 61: Can this be done in a more go-like way without an infinite loop here? What are we looping over?

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
 	lookup, _ := s.Factory.Subfactory.MakeLookup()
 	var nameserver string
```
zakird, line 57: These variables don't look like they're used outside of the loop. Is there a reason they can't just be defined within the loop?

modules/zone/zone.go
zakird: In all honesty, I can't immediately figure out why this loop is here at all and there's nearly 100 lines of code in the loop, both of which likely indicate that this code should be restructured. Further, it seems like this loop is busy waiting? Granted, I'm not really sure on what.

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
+			s.Factory.Factory.CHmu.Unlock()
+		} else {
+			waiting = tmp.([]*dns.Token)
+			waiting = append(waiting, record)
+			s.Factory.Factory.Hash.Add(domain, waiting)
+			s.Factory.Factory.CHmu.Unlock()
+			return nil, zdns.STATUS_NO_OUTPUT, nil
+		}
+		//First pass looking for a nameserver we know the IP for
+		var result interface{}
```
zakird, line 90: why are these being defined up here again?

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
-	s.Factory.Factory.GlueLock.RUnlock()
-	//Second pass performing lookups to find a nameserver
-	s.Factory.Factory.GlueLock.Lock()
-	if _, ok := (*s.Factory.Factory.Glue)[nameserver]; !ok {
-		result, status, err = lookup.(*alookup.Lookup).DoLookup(nameserver[0 : len(nameserver)-1])
-		if status == zdns.STATUS_NOERROR && err == nil {
-			addresses := append(result.(alookup.Result).IPv4Addresses, result.(alookup.Result).IPv6Addresses...)
-			(*s.Factory.Factory.Glue)[nameserver] = addresses
-			result, status, err = LookupHelper(domain, addresses, lookup.(*alookup.Lookup))
+		if domain[len(domain)-1] == '.' {
```
zakird, line 73: Is there a reason to not just use `strings.TrimSuffix` here?

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
-		for unsolved := range notify {
-			if unsolved {
-				proceed = true
-				break
-			} else {
-				notify <- false
-				break
-			}
+	for {
+		switch typ := record.RR.(type) {
```
zakird, line 62: This should either be an if statement or a type assertion. It doesn't make sense to have a case statement with one option.

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
+		case *dns.NS:
+			nameserver = strings.ToLower(typ.Ns)
+			domain = record.RR.Header().Name
+			domain = strings.Join([]string{prefix, domain}, "")
```
zakird, line 66: Is prefix expected to have a period by the other modules?

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
-	var status zdns.Status
-	var err error
-	s.Factory.Factory.GlueLock.RLock()
-	if locations, ok := (*s.Factory.Factory.Glue)[nameserver]; ok {
-		result, status, err = LookupHelper(domain, locations, lookup.(*alookup.Lookup))
-		if status == zdns.STATUS_NOERROR && err == nil {
-			s.Factory.Factory.GlueLock.RUnlock()
-			notify <- false
-			return result, status, err
+		if strings.Count(record.RR.Header().Name, ".") < 2 {
```
zakird, line 70: What's being checked here?

modules/zone/zone.go
```diff
@@ -189,52 +156,44 @@ func (s *RoutineLookupFactory) MakeLookup() (zdns.Lookup, error) {
...
-	CacheFactor int
-	CacheHash   *cachehash.CacheHash
-	CHmu        sync.Mutex
-	Manager     CallbackManager
+	Glue         *map[string][]string
+	GlueLock     sync.Mutex
+	Subfactory   alookup.GlobalLookupFactory
+	CacheFactor  int
+	GlueFilePath string
+	Hash         *cachehash.CacheHash
```
zakird, line 164: What is this a Hash of? What is it used for? Generally the type of a variable is not a good name for it.

modules/zone/zone.go
```diff
@@ -189,52 +156,44 @@ func (s *RoutineLookupFactory) MakeLookup() (zdns.Lookup, error) {
...
+	GlueCache    *cachehash.CacheHash
+	CHmu         sync.Mutex
 }
 
 func (s *GlobalLookupFactory) AddFlags(f *flag.FlagSet) {
 	f.BoolVar(&s.Subfactory.IPv4Lookup, "ipv4-lookup", true, "perform A lookups for each server")
 	f.BoolVar(&s.Subfactory.IPv6Lookup, "ipv6-lookup", true, "perform AAAA record lookups for each server")
-	f.IntVar(&s.CacheFactor, "cache-size-factor", 25, "number of times larger than the number of threads to make the cache of successes")
+	f.IntVar(&s.CacheFactor, "cache-size-factor", 25, "number of times larger than the number of threads to make the cache for looked up nameservers")
+	f.StringVar(&s.GlueFilePath, "glue-file", "", "glue file path")
```
zakird, line 173: I thought we were doing two passes over the original file instead of requiring it to be passed in twice. What happened to that approach?

modules/zone/zone.go
zakird: Why is `lookup` defined here? This isn't used until L103, and there are exit branches in the function between here and then.

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
 	lookup, _ := s.Factory.Subfactory.MakeLookup()
 	var nameserver string
 	var domain string
-	var notify chan bool
+	var waiting []*dns.Token
 	prefix := s.Factory.Factory.GlobalConf.NamePrefix
```
zakird, line 60: Same question as above. This is only used once in the function inside of an if statement. Why is it being defined here?

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
-		if status == zdns.STATUS_NOERROR && err == nil {
-			addresses := append(result.(alookup.Result).IPv4Addresses, result.(alookup.Result).IPv6Addresses...)
-			(*s.Factory.Factory.Glue)[nameserver] = addresses
-			result, status, err = LookupHelper(domain, addresses, lookup.(*alookup.Lookup))
+		if domain[len(domain)-1] == '.' {
+			domain = domain[0 : len(domain)-1]
+		}
+		s.Factory.Factory.CHmu.Lock()
+		tmp, found := s.Factory.Factory.Hash.Get(domain)
+		if !found {
```
zakird, line 78: This logic needs to be explained somewhere. It's not obvious what's going on here.

modules/zone/zone.go
```diff
@@ -93,82 +56,86 @@ func (s *Lookup) DoZonefileLookup(record *dns.Token) (interface{}, zdns.Status,
...
+			waiting = []*dns.Token{}
+			s.Factory.Factory.Hash.Add(domain, waiting)
+			s.Factory.Factory.CHmu.Unlock()
+		} else {
+			waiting = tmp.([]*dns.Token)
+			waiting = append(waiting, record)
+			s.Factory.Factory.Hash.Add(domain, waiting)
+			s.Factory.Factory.CHmu.Unlock()
+			return nil, zdns.STATUS_NO_OUTPUT, nil
```
zakird, line 87: This is likely due to not being able to understand this logic, but why is `nil` being returned here?

zakird (changes requested):
I started going through this code in depth, but as I went, I'm mostly coming to the conclusion that this code probably needs to be restructured. It doesn't make a lot of sense in it's current state and it's very difficult to understand what's going on.

zakird:
Given that there hasn't been any progress on this, I'm going to close this pull request. I'm probably going to pull zone record processing out of ZDNS for version 1.0 since the code is in poor shape and I'd like to focus on fixing other issues (e.g., fixing ALOOKUP module). We can ship it in a future release once it's been rewritten.

PR #77, opened 2016-10-10
Alookup

zakird:


zakird:
This should fix the bad logic in the `alookup` module.

umbernhard (commented):
Looks alright to me, thought admittedly I'm probably not the right person to be making this call.

umbernhard:
Also you'll have to pull master before you merge

modules/alookup/alookup.go
```diff
@@ -40,55 +41,71 @@ func (s *Lookup) DoLookup(name string) (interface{}, zdns.Status, error) {
 	return s.DoTargetedLookup(name, nameServer)
 }
 
-func (s *Lookup) DoTargetedLookup(name, nameServer string) (interface{}, zdns.Status, error) {
-	requests := []uint16{}
-	if s.Factory.Factory.IPv6Lookup {
-		requests = append(requests, dns.TypeAAAA)
+func (s *Lookup) doLookupProtocol(name string, nameServer string, dnsType uint16, searchSet *map[string][]miekg.Answer, origName string, depth int) ([]string, zdns.Status, error) {
```
dadrian: Don't pass a map by pointer, just pass it by value. Maps are reference values. See http://stackoverflow.com/questions/14928826/passing-pointers-to-maps-in-golang

dadrian (changes requested):

dadrian (approved):

PR #79, opened 2016-12-24
Support IPv6 nameservers in resolv.conf & zone files

hdm:
This fixes the following error when IPv6 nameservers are found in the resolv.conf file or zone files:
```
{"name":"example.com","status":"ERROR","error":"dial udp: too many colons in address 2001:4860:4860::8888:53"}
```

For command-line specification of IPv6 nameservers, these need to have the brackets applied manually, since otherwise it would difficult to disambiguate the address from the port:
```
-nameservers [2001:4860:4860::8888]:53
```

### DIFF
--- a/lookup.go
+++ b/lookup.go
@@ -39,7 +39,7 @@ func GetDNSServers(path string) ([]string, error) {
 	}
 	var servers []string
 	for _, s := range c.Servers {
-		if strings.Contains(s, ":") {
+		if s[0:1] != "[" && strings.Contains(s, ":") {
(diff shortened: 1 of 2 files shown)


dadrian (commented):
Arguably, we should validate addresses using `net.ParseIP`, but given that we're already just passing strings around, I don't see any reason to not accept this PR.

I'll still let @zakird or another primary author give final approval, but offhand LGTM!

Thanks for the contribution!

zakird (approved):

PR #83, opened 2017-01-05
Allow 'zdns -h' to work even without a lookup module

paul-pearce:
This is a mild hack. What it does is it adds flag parsing on the failure
path when no valid module is found. The flag parsing will find the "-h"
and print the default usage, terminating the program before hitting
the log.Fatal(). If no -h is supplied it falls through the flag parse
and hits the log.Fatal() as it did previously.

### DIFF
--- a/zdns/main.go
+++ b/zdns/main.go
@@ -67,6 +67,7 @@ func main() {
 	gc.Module = strings.ToUpper(os.Args[1])
 	factory := zdns.GetLookup(gc.Module)
 	if factory == nil {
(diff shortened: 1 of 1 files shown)


dadrian (approved):
I believe the proper way to do this is to push to stable, then PR to master.

Also, you won't believe how long it took me to explain this PR to @zakird. I could write a clickbait article about it.

dadrian:
> You won't believe how this Iranian hacker was thwarted by a grad student!

dadrian:
Thank you for your contribution!

Issue #85, opened 2017-01-09
Change default timeout value


paul-pearce:
Currently the default zdns timeout value is 10 seconds. The default BIND timeout value is also 10 seconds. This creates a situation where zdns frequently returns `TIMEOUT` a few milliseconds before BIND returns `SERVFAIL` (assuming default invocations of both).

Bumping this to 15 seconds seems reasonable.

dadrian:
Why not 11 seconds?

paul-pearce:
Sure. I was going for a value that seemed "normal" by human standards, since it doesn't actually matter. Assuming a properly configured upstream resolver (which is totally trivial, right @zakird?) 11s should behave the same as 15s, over a large number of queries.

PR #87, opened 2017-01-09
Change default timeout value from 10s to 15s

paul-pearce:
The BIND default timeout value is 10s. If zdns also uses a value
of 10s you get zdns timeouts when a few ms later BIND returns a
SERVFAIL. Depending on thread scheduling etc, authoritative
timeouts become a mix of TIMEOUT and SERVFAIL. This results
in overall sadness.


zakird (approved):

paul-pearce:
Fixes #85

PR #89, opened 2017-01-10
Add the option of including a timestamp in every output json

paul-pearce:
Adds `--timestamp` which will include a golang time.Time in every
output object. If ``--timestamp` is not set, the field is omitted.

Usage:
```
$ echo "example.com" | ./zdns A
{"name":"example.com","status":"NOERROR","data":{"answers":[{"ttl":49854,"type":"A","name":"example.com","answer":"93.184.216.34"}],"additionals":[],"authorities":[],"protocol":"udp"}}
```

```
$ echo "example.com" | ./zdns A --timestamp
{"name":"example.com","status":"NOERROR","timestamp":"2017-01-09T17:05:54.706439267-08:00","data":{"answers":[{"ttl":49626,"type":"A","name":"example.com","answer":"93.184.216.34"}],"additionals":[],"authorities":[],"protocol":"udp"}}
```

Let me answer some questions you're going to ask.

* Paul, why did you make `Timestamp` of type `*time.Time` instead of `time.Time`
  * Because `time.Time` can not be `empty`, meaning `omitempty` does not work right
* Paul, why do you create a variable and then take it's address (`now`) instead of just taking the address of `time.Now()`
  * Because that goes against the spec, despite the fact that `time.Now()` can't return `nil`
* Paul,  why are you doing this in doLookup() instead of miekg.DoLookup(), which would be closer to the wire time. 
  * Because the config information is not passed that deep into the depths of zdns.

### DIFF
--- a/conf.go
+++ b/conf.go
@@ -51,7 +51,7 @@ type Result struct {
 	AlexaRank   int         `json:"alexa_rank,omitempty"`
 	Status      string      `json:"status,omitempty"`
 	Error       string      `json:"error,omitempty"`
-	Timestamp   time.Time   `json:"timestamp"`
+	Timestamp   string      `json:"timestamp,omitempty"`
 	Data        interface{} `json:"data,omitempty"`
 }
 
--- a/conf.go
+++ b/conf.go
@@ -17,12 +17,11 @@ package zdns
 import "time"
 
 type GlobalConf struct {
-	Threads      int
-	Timeout      time.Duration
-	AlexaFormat  bool
-	LogTimestamp bool
-	GoMaxProcs   int
-	Verbosity    int
+	Threads     int
+	Timeout     time.Duration
+	AlexaFormat bool
+	GoMaxProcs  int
+	Verbosity   int
 
 	NameServersSpecified bool
 	NameServers          []string
@@ -52,7 +51,7 @@ type Result struct {
 	AlexaRank   int         `json:"alexa_rank,omitempty"`
 	Status      string      `json:"status,omitempty"`
 	Error       string      `json:"error,omitempty"`
-	Timestamp   *time.Time  `json:"timestamp,omitempty"`
+	Timestamp   time.Time   `json:"timestamp"`
 	Data        interface{} `json:"data,omitempty"`
 }
 
--- a/conf.go
+++ b/conf.go
@@ -17,11 +17,12 @@ package zdns
 import "time"
 
 type GlobalConf struct {
-	Threads     int
-	Timeout     time.Duration
-	AlexaFormat bool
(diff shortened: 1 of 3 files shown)


zakird:
fwm. @dadrian please confirm this is teh right way of doing this in golang

paul-pearce:
Personally, I'd argue for this being enabled by default.

zakird:
I am not fundamentally opposed to that.

dadrian:
This seems like a fine way to handle it being optional. However, I see no reason not to make this default and non-configurable, which means we could just use a `time.Time` instead of a pointer.

zakird:
I'm alright with it just being included.

dadrian:
That being said, this is a different time format than every other Z* project.

zakird:
Can someone paste what it is? Is it not 8601?

dadrian:
`2017-01-09T17:05:54.706439267-08:00`

zakird:
What's the ZGrab one? That level of precision doesn't seem necessary
offhand.

dadrian:
ZGrab is `"timestamp": "2017-01-10T14:55:34-05:00"`

zakird:
I vote have zdns match zgrab

conf.go
```diff
@@ -51,6 +52,7 @@ type Result struct {
 	AlexaRank   int         `json:"alexa_rank,omitempty"`
 	Status      string      `json:"status,omitempty"`
 	Error       string      `json:"error,omitempty"`
+	Timestamp   *time.Time  `json:"timestamp,omitempty"`
 	Data        interface{} `json:"data,omitempty"`
 }
```
dadrian, line 56: Let's change the output format to match ZGrab.

If you're not familiar with how to do this "cleanish" in Golang, you're going to have to write a custom `func (r *Result) MarshalJSON() ([]byte, error)` function to handle the formatting. I'd probably put this in a new file called `json.go`. You'll need to create a private auxillary type that has a `Timestamp string`, and manually convert using `time.Format`. Then just `return json.Marshal(&auxStruct)`.

dadrian (changes requested):

paul-pearce:
Changes:

-Include timestamps by default
-Change to RFC3339 timestamps
-Store the the timestamp as RFC3339 instead of time.Time. I did this as a tradeoff to avoid a custom JSONMarshal. Given that the timestamp is generated in the same code block as the call to marshal, it seemed like overkill/overengineering to implement a custom marshaller to keep the true time.Time type for ~3 lines of code. If we need to use the timestamp in the future for anything other than printing, I'll volunteer to extend it with a custom marshaller.

dadrian (approved):

PR #94, opened 2017-03-30
Support for fully contained iterative dns resolutions (no name server necessary)

paul-pearce:
This code, much like the rest of zdns, is in a beta state. Comments and further testing are welcome.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.5.2
+- 1.6.2
```
dadrian, line 3: Should maybe also build against 1.7.x and 1.8.x?

integration_tests.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def test_a_lookup_iterative(self):
+        c = u"./zdns/zdns alookup --ipv4-lookup --ipv6-lookup --iterative"
+        name = u"www.zdns-testing.com"
+        cmd, res = self.run_zdns(c, name)
+        self.assertSuccess(res, cmd)
+        self.assertEqualALookup(res, self.ALOOKUP_WWW_ZDNS_TESTING)
+
+
+if __name__ == '__main__':
```
dadrian, line 233: What are these tests in Python and not in Go?

interface.go
```diff
@@ -85,7 +85,7 @@ type GlobalLookupFactory interface {
 	// Help text for the CLI
 	Help() string
 	// Return a single scanner which will scan a single host
-	MakeRoutineFactory() (RoutineLookupFactory, error)
+	MakeRoutineFactory(int) (RoutineLookupFactory, error)
```
dadrian, line 88: Add a comment explaining this int

zakird, line 88: what is this int??? This seems sketchy to me.

dadrian: Please run go lint on your code and fix some of the style errors, for example write comments following convention.

lookup.go
```diff
@@ -65,8 +65,8 @@ func makeName(name string, prefix string) (string, bool) {
 	}
 }
 
-func doLookup(g *GlobalLookupFactory, gc *GlobalConf, input <-chan interface{}, output chan<- string, metaChan chan<- routineMetadata, wg *sync.WaitGroup) error {
-	f, err := (*g).MakeRoutineFactory()
+func doLookup(g *GlobalLookupFactory, gc *GlobalConf, input <-chan interface{}, output chan<- string, metaChan chan<- routineMetadata, wg *sync.WaitGroup, threadID int) error {
+	f, err := (*g).MakeRoutineFactory(threadID)
```
dadrian, line 69: I don't think you need to deference this to call a method on it

modules/axfr/axfr.go
```diff
@@ -148,10 +147,11 @@ func (s *GlobalLookupFactory) AddFlags(f *flag.FlagSet) {
 	f.StringVar(&s.BlacklistPath, "blacklist-file", "", "blacklist file for servers to exclude from AXFR lookups")
 }
 
-func (s *GlobalLookupFactory) MakeRoutineFactory() (zdns.RoutineLookupFactory, error) {
+func (s *GlobalLookupFactory) MakeRoutineFactory(threadID int) (zdns.RoutineLookupFactory, error) {
```
zakird, line 150: This really seems like something that should be available as a global or something and not something that we have to be passing around. I remain _very_ skeptical about this.

zdns/main.go
```diff
@@ -98,13 +92,20 @@ func main() {
...
 	// complete post facto global initialization based on command line arguments
 	gc.Timeout = time.Duration(time.Second * time.Duration(*timeout))
+	gc.IterationTimeout = time.Duration(time.Second * time.Duration(*iterationTimeout))
+	gc.IterativeTimeout = time.Duration(time.Second * time.Duration(*iterativeTimeout))
 	if *servers_string == "" {
-		// figure out default OS name servers
-		ns, err := zdns.GetDNSServers(*config_file)
-		if err != nil {
-			log.Fatal("Unable to fetch correct name servers:", err.Error())
+		// if we're doing recursive resolution, figure out default OS name servers
```
dadrian, line 96: s/recursive/iterative

zdns/main.go
```diff
@@ -52,14 +40,20 @@ func main() {
...
 	flags.IntVar(&gc.Verbosity, "verbosity", 3, "log verbosity: 1 (lowest)--5 (highest)")
+	flags.IntVar(&gc.Retries, "retries", 1, "how many times should zdns retry query if timeout or temporary failure")
+	flags.IntVar(&gc.MaxDepth, "max-depth", 10, "how deep should we recurse when performing iterative lookups")
+	flags.IntVar(&gc.CacheSize, "cache-size", 10000, "how many items can be stored in internal recursive cache")
 	servers_string := flags.String("name-servers", "", "comma-delimited list of DNS servers to use")
 	config_file := flags.String("conf-file", "/etc/resolv.conf", "config file for DNS servers")
-	timeout := flags.Int("timeout", 15, "timeout for resolving an individual name")
+	timeout := flags.Int("timeout", 15, "timeout for resolving an individual name (unused if --iterative)")
+	iterationTimeout := flags.Int("iteration-timeout", 4, "timeout for resolving a single iteration in an iterative query")
+	iterativeTimeout := flags.Int("iterative-timeout", 15, "timeout for completely resolving an individual name, iteratively")
```
dadrian: I don't see why we need `--timeout` and `--iterative-timeout`. Why are these not just the same?

modules/nslookup/nslookup.go
```diff
@@ -121,41 +119,35 @@ func DoNSLookup(name string, nameServer string, client *dns.Client, tcpClient *d
...
+	return s.DoNSLookup(name, s.Factory.Factory.IPv4Lookup, s.Factory.Factory.IPv6Lookup)
 }
 
 // Per GoRoutine Factory ======================================================
 //
 type RoutineLookupFactory struct {
 	miekg.RoutineLookupFactory
-	Factory *GlobalLookupFactory
+	Factory  *GlobalLookupFactory
+	ThreadID int
```
zakird: Wait, why is this defined here and not in `miekg.RoutineLookupFactory`?

integration_tests.py
dadrian: Ok so, I understand writing some stuff in Python to actually call the binary we build, but why are there _no_ tests in Go?

dadrian (dismissed):

zdns/main.go
zakird: How is this different than normal timeout?

zakird:
I don't really like how this threadID is getting passed around. @dadrian thinks we should pass around the name we were originally trying to lookup. I think we should use something like http://blog.sgmansfield.com/2015/12/goroutine-ids/.

interface.go
paul-pearce: Deferring this to the threadID discussion below.

modules/axfr/axfr.go
paul-pearce: Deferring this to the threadID discussion below

modules/nslookup/nslookup.go
paul-pearce: Good catch. Removed.

.travis.yml
paul-pearce: I'll defer to you all on this decision. I simply rev'd to the version that's common across both our infrastructures.

integration_tests.py
paul-pearce: Deferring to Zakir as he setup the testing infrastructure.

interface.go
paul-pearce: The head of master has 220 golint errors. This branch reduces the number of golint errors down to 150. I suggest we tackle conforming to golint as a separate PR across the entire project.

lookup.go
paul-pearce: Is your thinking?

`f, err := g.MakeRoutineFactory(threadID)`

If so:
`
../lookup.go:69: g.MakeRoutineFactory undefined (type *GlobalLookupFactory is pointer to interface, not interface)
`

zdns/main.go
paul-pearce: Hmm. iterativeTimeout was intended to be set to 10 by default, to match bind functionality. That's my mistake.

As for why they are different: They are for fundamentally different functions. timeout's primary job is to govern interacting with BIND instances that have a specific timeout default (generally 10 seconds), so we set it >10. iterativeTimeout is more governing an iterative process that involves interacting with multiple servers.

I'd suggest we keep the 3 separate to allow us to, by default, mimic the BIND timeout behavior.

paul-pearce: See discussion above.

paul-pearce:
WRT ThreadID. I preferred the solution of passing the the ID context to constructors more than using something that begins "if you use this, you will go straight to hell."

@zakird can you clarify why you dislike passing the context to the constructors?

interface.go
dadrian: My point was to just try to hit some of the low hanging fruit, such as comment style because they're easy, not to imply you had to fix all the go lint errors.

dadrian (approved):
I'd like to see issues filed for:

1. Refactoring this to have some sort of context that's passed around in a sane way. I don't have a good sense for how the code is laid out currently, but given that it's so damn hard to track Goroutines, my sense is that something should be architected differently.

2. More things should be tested in Golang itself

3. We should clean up all the comments, so that the comments provide documentation suitable for godoc.org.

4. It seems like we're doing some non-Go-things, like passing pointers-to-interfaces around, instead of interfaces. We should clean those up.

But none of these are @paul-pearce's fault, at least within the scope of iterative resolution. I am just worried about ZDNS turning into a ZGrab project of Go debt.

I'll approve and leave the rest of the review to @zakird

zakird:
Given that @dadrian's comments are non-specific and largely unrelated to this branch, I'm mostly fine putting those aside for now. However, I still do not understand these timeout flags (even after reading @paul-pearce's response several times), and I don't want to merge until we have more helpful descriptions.

zakird:
@paul-pearce my concern with the threadID is that it's really ugly to have to pass an integer to _every_ function call in the entire application for debugging purposes.

zakird:
@paul-pearce when you're online, can you ping me, and I can try to better understand what these command line flags do.

zakird (changes requested):
Blocking until we can sort out command line flags.

integration_tests.py
zakird: While I agree that unit tests would be helpful in this code base, these provide coverage over core functionality, and there isn't any benefit of writing a test handler in go here instead of python. Go-based tests are not inherently better.

paul-pearce:
@zakird, re "ThreadID being in _every_ function call". I don't follow. It is only passed into the MakeRoutineFactory() constructors for each module. There's one logging function in miekg that it's also passed into, since it's called from a global context instead of a routine context. Is there confusion on the scope of the state, or language issues?

RE command line arguments. As discussed outside this thread, I'd personally like to see the following defaults:

timeout -> 15s, to interface with default bind timeout, without races
iterativeTimeout -> 10s, to match default bind
iterationTimeout -> 3s, to match default bind

I only broke the 3 out as separate options to emulate the desired 3 defaults (and then forget to reset them after testing, sigh).

But honestly I don't care enough to hold up the merge. If there's broad agreement, I'm fine with:

timeout -> 15s, used for both iterative and recursive timeouts
iterationTimeout -> 3 or 4s.

zakird:
I have fixed timeout. 

re "ThreadID being in every function call" => things like `GetCachedResult` now have to take a threadID. That's what I find incredibly ugly.

zakird (approved):

PR #96, opened 2017-04-06
fix #88 and documents how iteration works

### LINKED ISSUE #88: README has broken syntax
`cat top-1m.csv | zdns MX --lookup-ipv4 --alexa` all of the lookup-ipv4 commands are wrong. I believe they should be `ipv4-lookup`. But an overall check of the correctness of the readme should be performed.

zakird:


paul-pearce (approved):
lgtm

PR #97, opened 2017-04-09
Bring axfr back

zakird:


dadrian (approved):
http://www.reactiongifs.com/r/2013/01/didnt-read-lol-chicken-gif.gif

modules/axfr/axfr.go
```diff
@@ -163,6 +159,9 @@ func (s *GlobalLookupFactory) Initialize(c *zdns.GlobalConf) error {
 			return err
 		}
 	}
+	if c.IterativeResolution == true {
```
paul-pearce, line 162: Can we get more info about why this is true? I know it throws a segfault, can we paste into this PR (for future reference) a trace, etc?

PR #100, opened 2017-05-02
Restore support for PTR queries

paul-pearce:
1) Restores support for PTR queries. Yay.

2) Adds unit tests for PTR queries. Yay.

Turns out unit tests can't help you identify when support for an entire class of queries vanishes, if you also forget to write a unit test for that class of queries.

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -104,7 +104,7 @@ class Tests(unittest.TestCase):
       }
     }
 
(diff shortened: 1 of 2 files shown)


paul-pearce:
WRT if the logic is correct: I lifted the PTR logic code out of the commit where the PTR module was removed.

paul-pearce:
Addresses #99

zakird (approved):

Issue #102, opened 2017-05-16
Add option to configure DNS class to query


stklcode:
I recently had a setup where I had to query records in unusual classes, different from the default `INET` and therefore hardcoded the classes and built ZDNS with these changes.

Now I implemented an additional parameter `--class <CLASS>` which allows to specify the class beforehand. Possible values are `INET`, `CSNET`, `CHAOS`, `HESIOD`, `NONE`, `ANY` - defaults to `INET`. It obviously does not make sense in some combinations (maybe junk analysis), but it leaves the user all freedom for configuration.

This change does not introduce any overhead in the query itself. However to be consistent and do not loose any information (e.g. if `ANY` is queried), the class of both the query and the answer have been added to the resulting JSON, which increases the result size by 13 characters per record.

I'd like to see this addition in ZDNS, if it fits the scope.

zakird:
Closing this since #103 is merged. Thank you for this PR. We were commenting this morning how this was one of the best documented and executed PRs we've seen in a while. :)

PR #103, opened 2017-05-16
Add option to configure DNS class to query

stklcode:
Addresses issue #102


zdns/main.go
```diff
@@ -93,6 +95,24 @@ func main() {
 	// complete post facto global initialization based on command line arguments
 	gc.Timeout = time.Duration(time.Second * time.Duration(*timeout))
 	gc.IterationTimeout = time.Duration(time.Second * time.Duration(*iterationTimeout))
+	// class initialization
+	switch *class_string {
```
zakird: It would be nice to do a `strings.ToUpper()` here. Usability wise, it's a bit friendlier to also allow lowercase values.

zakird (approved):
Overall this looks great to me. I left one minor comment about accepting lowercase command line values.

zakird:
@paul-pearce would you be willing to take a second glance at this? It looks good to me.

stklcode:
Thanks. Just added the recommended case casting for the CLI parameter and also the common shotdhand values.

PR #81, opened 2017-01-04
Allow NIL tcp dns clients to miekg.DoLookup()

paul-pearce:
This enables a separate project I have to invoke miekg.DoLookup() directly.

PR #105, opened 2017-06-22
Re-expose the bottom-most doLookup behavior

paul-pearce:
Long ago we intentionally exposed this so other tools (Ahem, Paul's
stuff) can use this interface. Well, when Zakir and I refactored
miekg for Iterative lookups, we broke this. This once again
exposes the bottom most logic via DoLookupWorker (not a member).

(It also fixes a nameserver bug on an error branch.)


paul-pearce:
This tries to restore what was added in https://github.com/zmap/zdns/pull/81 and subsequently lost in https://github.com/zmap/zdns/pull/94

zakird (commented):
seems fine

zakird (approved):

PR #117, opened 2017-08-30
Fix Iteration

paul-pearce:
Turns out iteration was fundamentally broken, but in a way that worked the vast majority of the time.

We should add better iteration unit tests (will file a separate issue)


dadrian:
Noting the issue filed is #118

zakird (approved):

PR #124, opened 2017-09-02
update Sirupsen -> sirupsen

kumarde:
The Sirupsen package is lowercased (https://github.com/sirupsen/logrus/pull/384), which is what other parts of ZMap/Censys are using (https://github.com/censys/certificate-daemon).

For consistency (and to get code that uses zdns to build) I'm updating the import here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 zdns/zdns
 .*
+*.iml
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 zdns/zdns
 .*
+*.iml
```
paul-pearce, line 3: What is this for?

kumarde: IntelliJ project file

paul-pearce (approved):
lgtm

Issue #127, opened 2017-09-07
database storage not supported


xl7dev:
Not support Mongodb?

PR #129, opened 2017-09-11
Refactor input/output

thereallogani:
This allows domain inputs and result outputs to be handled in a more
modular way. Just add iohandlers to define what will populate the input
and output channels. This would allow a workaround to issues like #127.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"math/rand"
 	"strings"
+	"sync"
 
 	"github.com/miekg/dns"
 	log "github.com/sirupsen/logrus"
@@ -89,6 +90,22 @@ type GlobalLookupFactory interface {
 	RandomNameServer() string
 }
 
+// handle domain input
+type InputHandler interface {
+	// give the InputHandler access to the global config in case it needs any of the settings
+	Initialize(conf *GlobalConf)
+	// FeedChannel takes a channel to write domains to, the WaitGroup managing them, and if it's a zonefile input
+	FeedChannel(in chan<- interface{}, wg *sync.WaitGroup, zonefileInput bool) error
+}
+
+// handle output results
+type OutputHandler interface {
+	// give the OutputHandler access to the global config in case it needs any of the settings
+	Initialize(conf *GlobalConf)
+	// takes a channel (results) to write the query results to, and the WaitGroup managing the handlers
+	WriteResults(results <-chan string, wg *sync.WaitGroup) error
+}
+
 type BaseGlobalLookupFactory struct {
 	GlobalConf *GlobalConf
 }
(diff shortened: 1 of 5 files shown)


zakird (approved):
This looks good to me offhand. Asking @justinbastress to take a second look and make sure he's also cool with the change.

iohandlers/file/file.go
```diff
@@ -0,0 +1,85 @@
...
+	h.filepath = conf.OutputFilePath
+}
+
+func (h *OutputHandler) WriteResults(results <-chan string, wg *sync.WaitGroup) error {
+	var f *os.File
+	if h.filepath == "" || h.filepath == "-" {
+		f = os.Stdout
+	} else {
+		var err error
+		f, err = os.OpenFile(h.filepath, os.O_WRONLY|os.O_CREATE, 0666)
```
justinbastress: 0666? Any reason other users need write permission?

interface.go
```diff
@@ -89,6 +90,18 @@ type GlobalLookupFactory interface {
 	RandomNameServer() string
 }
 
+// handle domain input
+type InputHandler interface {
+	Initialize(conf *GlobalConf)
+	FeedChannel(in chan<- interface{}, wg *sync.WaitGroup, zonefileInput bool) error
```
justinbastress: Would be helpful to have the args described

interface.go
```diff
@@ -131,13 +144,33 @@ func (s *BaseGlobalLookupFactory) ZonefileInput() bool {
 // keep a mapping from name to factory
 var lookups map[string]GlobalLookupFactory
 
+// keep a mapping from name to input handler
+var inputs map[string]InputHandler
```
justinbastress: Since this is shared throughout the zdns module, maybe rename to `inputHandlers`/`outputHandlers` or something even more specific?

iohandlers/file/file.go
```diff
@@ -0,0 +1,85 @@
...
+	} else {
+		s := bufio.NewScanner(f)
+		for s.Scan() {
+			in <- s.Text()
+		}
+		if err := s.Err(); err != nil {
+			log.Fatal("input unable to read file", err)
+		}
+	}
+	close(in)
```
justinbastress: defer `close`/`Done`, in case there is an unexpected panic?

iohandlers/file/file.go
justinbastress: This is 0644 in the old `doOutput` -- any reason this needs to be writable by everyone?

iohandlers/file/file.go
```diff
@@ -0,0 +1,85 @@
...
+		f, err = os.OpenFile(h.filepath, os.O_WRONLY|os.O_CREATE, 0666)
+		if err != nil {
+			log.Fatal("unable to open output file:", err.Error())
+		}
+		defer f.Close()
+	}
+	for n := range results {
+		f.WriteString(n + "\n")
+	}
+	(*wg).Done()
```
justinbastress: Does `(*wg).Done()` need to be deferred in case of an unexpected panic? This seems to match the current behavior, but that doesn't necessarily mean it's right.

justinbastress (changes requested):
I think the only critical changes would be the 0644->0666 mode, and ensuring that the streams get closed / signalled properly even in case of a panic.

justinbastress (approved):

Issue #122, opened 2017-09-01
Add `--nanoseconds` to shift from RFC3339 -> RFC3339Nano


paul-pearce:
The option in main should set one of these two constants from `time` in the config object, which every call to time.Now() should use (it uses the time.RFC3339 constant now).

PR #132, opened 2017-09-20
Output configuration data in metadata and add --nanoseconds

### LINKED ISSUE #120: GlobalConf (conf.go) should be included in the metadata file
Right now the metadata file is quite sparse and does not include a variety of information it really should. This can be solved most simply by exporting GlobalConf.

Some notes:

* I suspect the simplest way to do this is to put `c` into `metaChan` before calling `aggregateMetadata` https://github.com/zmap/zdns/blob/master/lookup.go#L230

* All of the metadata logic from `DoLookups` should be inside the if statement that checks if we're outputting metadata. We currently do a bunch of work constructing the metadata object and then throw it away if the flag is not set

paul-pearce:
Resolves #120 and #122.

### DIFF
--- a/conf.go
+++ b/conf.go
@@ -51,6 +51,7 @@ type Metadata struct {
 	NameServers []string       `json:"name_servers"`
 	Timeout     int            `json:"timeout"`
 	Retries     int            `json:"retries"`
+	Conf        *GlobalConf    `json:"conf"`
 }
 
 type Result struct {
--- a/conf.go
+++ b/conf.go
@@ -28,6 +28,7 @@ type GlobalConf struct {
 	CacheSize           int
 	GoMaxProcs          int
 	Verbosity           int
+	TimeFormat          string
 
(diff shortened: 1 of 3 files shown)


dadrian (commented):
Do we want to _not_ output `metadata-file` if it's not passed, or default to `stderr`?

zakird (approved):

paul-pearce:
@dadrian We currently do not output metadata at all if `metadata-file` is not set (we do not send to `stderr`). I'd be happy to discuss this in an issue if you feel strongly, but I _really_ want zdns to replace dig for my daily life, and if I have to pipe stderr around by default every time I use it I'm going to be significantly less happy with my daily life.

zakird:
I agree with Paul on this one. Especially given that we're also going to be making `--short` the new default output.

dadrian:
Sounds reasonable to me!

PR #136, opened 2017-10-09
Allow `dig like` resolution of a single name from the cmd line.

paul-pearce:
Usage:
```
$ zdns A example.com | jq "."
{
  "data": {
    "flags": {
      "error_code": 0,
      "response": true,
      "opcode": 0,
      "authoritative": false,
      "truncated": false,
      "recursion_desired": true,
      "recursion_available": true,
      "authenticated": false,
      "checking_disabled": false
    },
    "protocol": "udp",
    "authorities": [],
    "additionals": [],
    "answers": [
      {
        "answer": "93.184.216.34",
        "name": "example.com",
        "class": "IN",
        "type": "A",
        "ttl": 63020
      }
    ]
  },
  "timestamp": "2017-10-08T18:20:45-07:00",
  "status": "NOERROR",
  "class": "IN",
  "name": "example.com"
}
```

### DIFF
--- a/lookup.go
+++ b/lookup.go
@@ -158,36 +158,42 @@ func doOutput(out <-chan string, path string, wg *sync.WaitGroup) error {
 
 // read input file and put results into channel
 func doInput(in chan<- interface{}, path string, wg *sync.WaitGroup, zonefileInput bool, passedName string) error {
+
+	defer func() {
+		close(in)
+		(*wg).Done()
+	}()
+
 	if len(passedName) > 0 {
 		in <- passedName
+		return nil
+	}
+
+	var f *os.File
+	if path == "" || path == "-" {
+		f = os.Stdin
 	} else {
-		var f *os.File
-		if path == "" || path == "-" {
-			f = os.Stdin
-		} else {
-			var err error
-			f, err = os.Open(path)
-			if err != nil {
-				log.Fatal("unable to open input file:", err.Error())
-			}
-		}
-		if zonefileInput {
-			tokens := dns.ParseZone(f, ".", path)
-			for t := range tokens {
-				in <- t
-			}
-		} else {
-			s := bufio.NewScanner(f)
-			for s.Scan() {
-				in <- s.Text()
-			}
-			if err := s.Err(); err != nil {
-				log.Fatal("input unable to read file", err)
-			}
+		var err error
+		f, err = os.Open(path)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


lookup.go
```diff
@@ -157,29 +157,33 @@ func doOutput(out <-chan string, path string, wg *sync.WaitGroup) error {
 }
 
 // read input file and put results into channel
-func doInput(in chan<- interface{}, path string, wg *sync.WaitGroup, zonefileInput bool) error {
-	var f *os.File
-	if path == "" || path == "-" {
-		f = os.Stdin
+func doInput(in chan<- interface{}, path string, wg *sync.WaitGroup, zonefileInput bool, passedName string) error {
+	if len(passedName) > 0 {
```
dadrian: Rather than a big `if...else` I'd probably put a `defer func() { close(in) }()` at the top, and then just have the `if len(passedName) > 0` block return.

lookup.go
```diff
@@ -157,29 +157,33 @@ func doOutput(out <-chan string, path string, wg *sync.WaitGroup) error {
...
+			s := bufio.NewScanner(f)
+			for s.Scan() {
+				in <- s.Text()
+			}
+			if err := s.Err(); err != nil {
+				log.Fatal("input unable to read file", err)
+			}
 		}
 	}
 	close(in)
```
dadrian: And then I'd remove this line.

zdns/main.go
```diff
@@ -144,6 +144,17 @@ func main() {
...
 		runtime.GOMAXPROCS(gc.GoMaxProcs)
 	}
+	if len(flags.Args()) > 0 {
+		stat, _ := os.Stdin.Stat()
+		// If stdin is piped from the terminal, and we havent specified a file, and if we have unparsed args
+		// use them for a dig like reslution
+		if (stat.Mode()&os.ModeCharDevice) != 0 && gc.InputFilePath == "-" && len(flags.Args()) == 1 {
+			gc.PassedName = flags.Args()[0]
+		} else {
+			log.Fatal("Unpassed command line flags: ", flags.Args())
```
dadrian: Shouldn't this be `unused`?

dadrian (approved):
Just some thoughts. Do with them what you will.

paul-pearce:
I intend to make the changes above. I don't have cycles at the moment. Please don't merge this PR until I do.

zakird:
looks fine to me.

Issue #141, opened 2017-10-21
NXDOMAINs result in STATUS_ERROR with --iterative


thereallogani:
Running something like `echo "imfakenotreal.com" | zdns A --iterative --verbosity 5` will return
```
log output...

DEBU[0000] THREADID 000004,DEPTH 02:        [-> Authority found, iterating]
DEBU[0000] THREADID 000004,DEPTH 02:        [Trying Authority:  {172800 NS 2 IN 1 com c.gtld-servers.net}]
DEBU[0000] THREADID 000004,DEPTH 02:        [Glue hit for Authority:  c.gtld-servers.net .  {172800 A 1 IN 1 c.gtld-servers.net 192.26.92.30}]
DEBU[0000] THREADID 000004,DEPTH 02:        [Output from extract authorities:  192.26.92.30:53]
DEBU[0000] THREADID 000004,DEPTH 02:        [iterative lookup for  imfakenotreal.com  ( 1 ) against  192.26.92.30:53  layer  com]
DEBU[0000] THREADID 000004,DEPTH 03:            [Cached retrying lookup. Name:  imfakenotreal.com , Layer:  com , Nameserver:  192.26.92.30:53]
DEBU[0000] THREADID 000004,DEPTH 04:                [Cache request for:  imfakenotreal.com  ( 1 )]
DEBU[0000] THREADID 000004,DEPTH 05:                    [-> no entry found in cache]
DEBU[0000] THREADID 000004,DEPTH 04:                [Cache auth check for  imfakenotreal.com]
DEBU[0000] THREADID 000004,DEPTH 05:                    [Cache request for:  imfakenotreal.com  ( 2 )]
DEBU[0000] THREADID 000004,DEPTH 06:                        [-> no entry found in cache]
DEBU[0000] THREADID 000004,DEPTH 04:                [Wire lookup for name:  imfakenotreal.com  ( 1 ) at nameserver:  192.26.92.30:53]
DEBU[0000] THREADID 000004,DEPTH 01:    [****WIRE LOOKUP*** imfakenotreal.com   192.26.92.30:53]
DEBU[0000] THREADID 000004,DEPTH 03:            [-> error occurred during lookup]
DEBU[0000] THREADID 000004,DEPTH 03:            [--> Auth resolution of  192.26.92.30:53  Failed:  NXDOMAIN]

more log messages...

{
  "name": "imfakenotreal.com",
  "class": "IN",
  "status": "ERROR",
  "error": "could not find authoritative name server",
[...]

PR #142, opened 2017-10-21
Check for NXDOMAINs rather than assuming a STATUS_ERROR

thereallogani:
This will address issue #141


paul-pearce:
This looks good. Thanks for finding this and the fix.

paul-pearce (approved):
lgtm

PR #146, opened 2017-11-15
Add handling for all zdns statuses

### LINKED ISSUE #144: Issue handling certain statuses
I've noticed some other issues similar to #141 but with other statuses such as SERVFAIL/REFUSED/etc. The verbose log shows what is really happening but the status doesn't get handled and the output status is just ERROR

thereallogani:
Add more handling of statuses to fix #144


zakird (approved):
This looks fine to me offhand. Asking @justinbastress to take a second look.

justinbastress (approved):

PR #147, opened 2017-11-15
Add DS/DNSKEY support

stirnim:
Adds lookup type DS and DNSKEY which are used in DNSSEC signed zones.


zakird (approved):

zakird:
Thank you for the PR, and again, my apologies for the delay in getting this merged back into mainline.

Issue #149, opened 2018-04-24
Integration tests failing


thereallogani:
The nameservers for the test domain (zdns-testing.com) appear to have changed from "ns-cloud-b?.googledomains.com" to "ns-cloud-c?.googledomains.com". Also the serial number on the SOA record appears to have changed as well (1 -> 2).

`echo 'zdns-testing.com' | zdns NS --iterative | jq .`
`{
  "name": "zdns-testing.com",
  "class": "IN",
  "status": "NOERROR",
  "timestamp": "2018-04-24T11:06:55-04:00",
  "data": {
    "answers": [
      {
        "ttl": 172800,
        "type": "NS",
        "class": "IN",
        "name": "zdns-testing.com",
        "answer": "ns-cloud-c1.googledomains.com"
      },
      {
        "ttl": 172800,
        "type": "NS",
        "class": "IN",
        "name": "zdns-testing.com",
        "answer": "ns-cloud-c2.googledomains.com"
      },
      {
        "ttl": 172800,
        "type": "NS",
        "class": "IN",
        "name": "zdns-testing.com",
        "answer": "ns-cloud-c3.googledomains.com"
      },
      {
        "ttl": 172800,
        "type": "NS",
        "class": "IN",
        "name": "zdns-testing.com",
        "answer": "ns-cloud-c4.googledomains.com"
      }
    ],
    "additionals": [],
    "authorities": [],
    "protocol": "",
    "flags": {
      "response": false,
      "opcode": 0,
      "authoritative": false,
      "truncated": false,
      "recursion_desired": false,
      "recursion_available": false,
      "authenticated": false,
      "checking_disabled": false,
      "error_code": 0
    }
  }
}
`

`echo 'zdns-testing.com' | zdns SOA --iterative | jq .`
`{
  "name": "zdns-testing.com",
  "class": "IN",
  "status": "NOERROR",
  "timestamp": "2018-04-24T11:08:00-04:00",
  "data": {
    "answers": [
      {
        "ttl": 21600,
        "type": "SOA",
        "class": "IN",
        "name": "zdns-testing.com",
        "ns": "ns-cloud-c1.googledomains.com",
        "mbox": "cloud-dns-hostmaster.google.com",
        "serial": 2,
        "refresh": 21600,
        "retry": 3600,
        "expire": 259200,
        "min_ttl": 300
      }
    ],
    "additionals": [],
    "authorities": [],
    "protocol": "udp",
    "flags": {
      "response": true,
[...]

PR #150, opened 2018-04-24
Fix tests

thereallogani:
#149 
Namerservers for the test domain seem to have changed from
ns-cloud-b?.googledomains.com to ns-cloud-c?.googledomains.com. Also the
serial number of the SOA answer appears to have changed as well


zakird (approved):
Thanks for catching this! I changed a configuration in Google the other day, but didn't realize it would break tests.

PR #153, opened 2018-08-03
Omits empty and nil "servers" list from result.

clayzim:
A nil sever field is invalid, so it should either not exist or be a list with members.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.7.4
+- 1.9.3
 before_install:
 - go get github.com/sirupsen/logrus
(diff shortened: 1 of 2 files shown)


clayzim:
Travis seems to only be failing due to `go get github.com/miekg/dns` failing.

justinbastress:
If I had to guess, this is probably a golang version issue...@zakird, does zdns need to work with older versions of go, or can we bump this?

zakird:
Bump it

justinbastress (approved):
Looks good to me.

dadrian (approved):

PR #154, opened 2018-09-05
Add blacklist to the miekg module

thereallogani:
This allows the same blacklist capability allowed by the axfr module. 

There were also some minor gofmt changes.

### DIFF
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -2,6 +2,7 @@ package miekg
 
 import (
 	"errors"
+	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


zakird:
I think this is fine, though worth noting it only really matters when zdns is doing local recursion since a recursive resolver might do as it pleases.

zakird:
@paul-pearce do you have any thoughts?

zakird (approved):

thereallogani:
@zakird I added a comment that makes clear this will only take effect for iterative queries

PR #166, opened 2019-08-19
do not omit zero TTL values in answer section (fix #165)

### LINKED ISSUE #165: TTL=0 does not get written to JSON output
Example of broken domain:
```
$ dig jucesc.sc.gov.br NS

; <<>> DiG 9.14.4 <<>> jucesc.sc.gov.br NS
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 19089
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 5

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;jucesc.sc.gov.br.		IN	NS

;; ANSWER SECTION:
jucesc.sc.gov.br.	0	IN	NS	srvjucesc01.jucesc.sc.gov.br.
jucesc.sc.gov.br.	0	IN	NS	ns1.sc.gov.br.
jucesc.sc.gov.br.	0	IN	NS	srvjucesc02.jucesc.sc.gov.br.
```

Output from zdns:
```
        "answers": [
            {
                "type": "NS",
                "class": "IN",
                "name": "jucesc.sc.gov.br",
                "answer": "ns1.sc.gov.br"
            },
            {
                "type": "NS",
                "class": "IN",
                "name": "jucesc.sc.gov.br",
                "answer": "srvjucesc01.jucesc.sc.gov.br"
            },
            {
                "type": "NS",
                "class": "IN",
                "name": "jucesc.sc.gov.br",
                "answer": "srvjucesc02.jucesc.sc.gov.br"
            }
        ],
```

This is unexpected and requires special handling in JSON reader. It would be much easier to always write TTL value, even if it is zero.

Thank you for considering this.

stklcode:
TTL is a mandatory value of the RR header and zero is a valid value.
However the Go JSON marshaller cannot distinguish between unset and zero numeric values so the result got dropped.

From what I think valid records with missing TTL (not zero, but completely missing) would be completely omitted in previous steps so the simple solution of not omitting "empty" (i.e. zero) values in the output should be OK.

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -137,7 +137,7 @@ class Tests(unittest.TestCase):
         "type":"PTR",
         "class":"IN",
         "name":"8.8.8.8.in-addr.arpa",
-        "answer":"google-public-dns-a.google.com."
+        "answer":"dns.google."
       }
     ]
 
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -18,7 +18,7 @@ import (
 )
 
 type Answer struct {
-	Ttl     uint32 `json:"ttl,omitempty"`
+	Ttl     uint32 `json:"ttl"`
 	Type    string `json:"type,omitempty"`
 	rrType  uint16
 	Class   string `json:"class,omitempty"`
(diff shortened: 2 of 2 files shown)


zakird (approved):
agree. thanks for fixing.

zakird:
Can you fix the corresponding test? Once that's done, I'm happy to merge.

stklcode:
It not a "corresponding" test that fails. Apparently Google has changed the PTR of 8.8.8.8 from "google-public-dns-a.google.com" to "dns.google" recently...

Fixed it anyway. (probably should not be squashed during merge as the changes are not related to the actual issue)

zakird:
Thanks for taking care of that.

Issue #163, opened 2019-06-23
alookup returns non ipv4 answers 


nikitaborisov:
It seems to happen when a DNS server returns records of different types in the answer section to an `A` query:

```json
$ echo s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net | ~/go/bin/zdns a -iterative
{"name":"s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net","class":"IN","status":"NOERROR","timestamp":"2019-06-23T14:59:09+02:00","data":{"answers":[{"ttl":3600,"type":"A","class":"IN","name":"s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net","answer":"95.85.16.212"},{"ttl":3600,"type":"AAAA","class":"IN","name":"s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net","answer":"2a03:b0c0:0:1010::509:d001"}],"additionals":[],"authorities":[],"protocol":"udp","flags":{"response":true,"opcode":0,"authoritative":false,"truncated":false,"recursion_desired":false,"recursion_available":true,"authenticated":false,"checking_disabled":false,"error_code":0}}}
$ echo s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net | ~/go/bin/zdns alookup -iterative
{"name":"s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net","class":"IN","status":"NOERROR","timestamp":"2019-06-23T14:59:56+02:00","data":{"ipv4_addresses":["95.85.16.212","2a03:b0c0:0:1010::509:d001"]}}
```

Note that the second address in the `ipv4_addresses` field of `data` is actually an ipv6 address, returned as an `AAAA` record in the DNS query. I've also seen a server name show up in the `ipv4_addresses` field though I can't reproduce it right now.

stklcode:
Apparently the nameserver (95.85.16.212 in this case) returns both A and AAAA in the answer section independent of the actual question and the tool (you might even try something weired like `s7hscpw1zwk8c8breqsiv6oiap07p6qz3dm0u5e3.ipleak.net. HS NAPTR` and will receive the `IN A`/`IN AAAA` tuple.

[...]

PR #167, opened 2019-08-21
add class filter for alookup module

stklcode:
Targeting #163 the alookup module currently returns all answers from the nameserver if the first answer matches the requested type (`A`/`AAAA`).

I've added a filter in the loop to only collect the requested type and `CNAME` records.

Because this still failed to deliver correct results if both an `A` and a `CNAME` record (never seen in the wild but you never know) or multiple `CNAME` records (which is not valid, but possible to configure with older BIND servers up to 9.1.0 <sup>1</sup>). In both cases the resulting domain names have been added to the IP set.

To avoid double filtering I've slightly refactored the routing to collect target type and CNAME in separate sets.

All changes covered by previously written tests.

----
[1] https://docstore.mik.ua/orelly/networking_2ndEd/dns/ch10_07.htm

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+import copy
 import subprocess
 import json
 import unittest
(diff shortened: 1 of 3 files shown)


zakird (approved):
Looks good. Thanks for the fix.

Issue #164, opened 2019-07-25
Some IPv6 addresses in AAAA printed using IPv4 syntax


pspacek:
Example on live Internet:
```
echo ns1.exporterapp.com  | zdns NS -trace -iterative
```

Broken fragment of output:
```
        {
            "cached": false,
            "class": 1,
            "depth": 3,
            "layer": "exporterapp.com",
            "name": "ns1.exporterapp.com",
            "name_server": "5.9.101.204:53",
            "results": {
                "additionals": [
                    {
                        "answer": "5.9.101.204",
                        "class": "IN",
                        "name": "ns1.exporterapp.com",
                        "ttl": 14400,
                        "type": "A"
                    },
                    {
                        "answer": "5.9.101.204",
                        "class": "IN",
                        "name": "ns1.exporterapp.com",
                        "ttl": 14400,
                        "type": "AAAA"
                    },
```

As far as I can see in Wireshark the server is answering correctly, it looks like problem in output formating.

Traditional dig prints this:
```
$ dig  ns1.exporterapp.com AAAA

; <<>> DiG 9.14.4 <<>> ns1.exporterapp.com AAAA
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 56865
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 4, ADDITIONAL: 4

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;ns1.exporterapp.com.		IN	AAAA

;; ANSWER SECTION:
ns1.exporterapp.com.	13855	IN	AAAA	::ffff:5.9.101.204

;; AUTHORITY SECTION:
ns1.exporterapp.com.	85855	IN	NS	ns1.exporterapp.com.
ns1.exporterapp.com.	85855	IN	NS	ns2.exporterapp.com.
ns1.exporterapp.com.	85855	IN	NS	ns4.exporterapp.com.
ns1.exporterapp.com.	85855	IN	NS	ns3.exporterapp.com.

;; ADDITIONAL SECTION:
ns2.exporterapp.com.	172255	IN	A	5.9.101.204
ns2.exporterapp.com.	13855	IN	AAAA	::ffff:5.9.101.204
ns1.exporterapp.com.	172255	IN	A	5.9.101.204

;; Query time: 3 msec
[...]

PR #168, opened 2019-08-21
add prefix to IPv4-mapped IPv6 addresses

stklcode:
Add a `::ffff:` prefix to IPv4-mapped IPv6 addresses when parsing AAAA answers.

Claimed in #164 mapped IPs like `::ffff:192.0.2.1` are stripped down `192.0.2.1` in ZDNS results.

Go's `net.IP` `String()` simply returns the 4-octet notation while the full 16 bytes are stored internally.
Trying to obtain 4-byte version suceeds on such addresses while it fails on "real" v6 addresses. In fact such constructs are used all over the `IP` lib:
```go
if ip4 := ip.To4(); ip4 != nil {
    // IPv4 address
} else {
    // IPv6 address
}
```

Using the same check we now add the well-known `::ffff:` prefix in those cases.

Started a unit test for the `ParseAnswer()` function to verify my changes.

### DIFF
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -154,7 +154,12 @@ func ParseAnswer(ans dns.RR) interface{} {
 			Name:    a.Hdr.Name,
 			Answer:  a.A.String()}
 	} else if aaaa, ok := ans.(*dns.AAAA); ok {
-		retv = Answer{Ttl: aaaa.Hdr.Ttl, Type: dns.Type(aaaa.Hdr.Rrtype).String(), rrType: aaaa.Hdr.Rrtype, Class: dns.Class(aaaa.Hdr.Class).String(), rrClass: aaaa.Hdr.Class, Name: aaaa.Hdr.Name, Answer: aaaa.AAAA.String()}
+		ip := aaaa.AAAA.String()
+		if aaaa.AAAA.To4() != nil {
+			// we have a IPv4-mapped address, append prefix (#164)
+			ip = "::ffff:" + ip
+		}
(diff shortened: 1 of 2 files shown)


stklcode:
Just checked what happens to a "IPv6-compatible address" (e.g. `::192.0.2.1`).

In fact Go does not even recognize such an address as a v4, so `ip.To4()` returns `nil` and `ip.String()` returns `::c000:201` in the above case.

There would be a solution:
```go
// verify we really got full 16-byte address
if len(aaaa.AAAA) == net.IPv6len {
	if aaaa.AAAA.To4() != nil {
		// we have a IPv4-mapped address, append prefix (#164)
		ip = "::ffff:" + ip
	} else {
		v4compat := true
		for _, o := range aaaa.AAAA[:11] {
			if o != 0 {
				v4compat = false
				break
			}
		}
		if v4compat {
			// we have a IPv4-compatible address, append prefix (#164)
			ip = "::" + aaaa.AAAA[12:].String()
		}
	}
}
```

~~**Question:**~~
Is there any need for that? ~~or can we assume this will never occur?~~
~~(I could not find a single record of this kind looking into an older .org zone scan)~~

**Edit:**
Rechecked the numbers above against a full `.org` zone scan from early 2017 (not collected with ZDNS):
* 4491 (329 unique) unique IPv4-mapped addresses (`::ffff:<IPv4>`)
* 222 (112 unique) IPv4-compatible addresses (`::<IPv4>`)

So yes, they do exist in the wild.

zakird (approved):
Looks good to me. Thanks for the investigation and fix.

PR #171, opened 2019-12-18
Adds --tcp-only and --udp-only flags

zakird:
Allow querying using only UDP or TCP.


dadrian (approved):
LGTM, but I'd remove `useTCP` because it's only read once and will match the state of `res.Protocol = "tcp"`, or is inferred by UDP or TCP object being non-null

dadrian (approved):

PR #172, opened 2019-12-20
use other miekg version that reuses sockets

zakird:


zakird:
closing since we're taking care of this in go modules.

Issue #63, opened 2016-08-25
Option to specify IP(s) from which we send queries


brandonpoc:
I have three local cacheing-only resolvers running 'unbound' and each resolver has 16 IPs attached to it (as well as ~1024 IPv6 IPs), through which unbound rotates round-robin when sending outbound requests on behalf of lookups made from local clients.

It would be nice if zdns had an option to specify from which IP(s) it sends its requests from (that is, which IP it would bind() to). In a case where I have multiple IPs on a machine, this would allow me to kind of spread the load around when doing heavy NS lookups on public servers. It could be presented with a CLI option like -source-ips IP1,IP2,IP3 etc like you may specify name servers. The purpose of this is to help avoid banning on public DNS servers during heavy query sessions. And, considering zdns was built for this kind of thing (heavy queries!), I'd imagine an option like this is a natural progression and should not be too difficult implement (in C I just set the bind() for the socket; with Go, I have no idea).

Thanks

zakird:
I agree that this would be potentially useful and fits the scope of ZDNS. However, given our workload, it's unlikely that the core team will have a chance to tackle this issue in the near future. if you have time to implement it, we'd be happy to accept a pull request.

emremuratercan:
Hello zakird i have an issue something like your wish. Did you handled with your problem?

zakird:
@stklcode fixed this in #173, which I merged with some additional commits.

PR #173, opened 2019-12-22
introduce --local-addr and --local-interface to set outgoing addresses

stklcode:
Addresses #63.

I've hardcoded such feature some time ago, but then moved to containerization to overcome this problem. In the meantime the libs have slightly changed so I came up with a simple solution to specify local addresses for the lookup.

Introducing new options `--local-addr` for a comma-separated list of local IPs (e.g. `--local-addr 192.0.2.1,192.0.2.2`) and `--local-interface` for a physical interface (e.g. `--local-interface enp0s1`) which reads all addresses from the given interface.

If both are specified, address takes precedence over interface as likely more specific.

On initialization of the lookup factories a random address is picked and the `Dialer` of the client is set.

Don't see any elegant way to add integration tests (don't know what interface(s) and address(es) are available at runtime) except binding to `127.0.0.1` with external nameserver and expecting failure.

**Points for discussion:**
In dual-stack environments there is currently no way to synchronize local and name server address. Giving an interface with both v4 and v6 addresses and a mixed list of nameservers the randomness of both selectors might raise an error like `no suitable address found`.

I personally think, misconfiguration is acceptable, so I didn't care about resolving potential mismatches yet.


zdns/main.go
```diff
@@ -59,6 +60,8 @@ func main() {
 	flags.BoolVar(&gc.TCPOnly, "tcp-only", false, "Only perform lookups over TCP")
 	flags.BoolVar(&gc.UDPOnly, "udp-only", false, "Only perform lookups over UDP")
 	servers_string := flags.String("name-servers", "", "comma-delimited list of DNS servers to use")
+	localaddr_string := flags.String("local-addr", "", "cp,,adeöo,oted öost pf local addresses to use")
```
zakird: Something got garbled here. I'm not sure what this is supposed to say.

zdns/main.go
```diff
@@ -139,6 +142,37 @@ func main() {
 		gc.NameServers = strings.Split(*servers_string, ",")
 		gc.NameServersSpecified = true
 	}
+	if *localaddr_string != "" {
+		for _, la := range strings.Split(*servers_string, ",") {
+			ip := net.ParseIP(la)
+			if ip != nil {
+				gc.LocalAddrs = append(gc.LocalAddrs, ip)
+				log.Info("using local address: ", localaddr_string)
```
zakird: This will print out multiple times, which doesn't seem correct, given that it will be printing out the entire CSV of the input.

zdns/main.go
```diff
@@ -139,6 +142,37 @@ func main() {
...
+			} else {
+				log.Fatal("Invalid argument for --local-addr (", la, "). Must be a comma-separted list of valid IP addresses.")
+			}
+		}
+		gc.LocalAddrSpecified = true
+	}
+	if *localif_string != "" {
+		if gc.LocalAddrSpecified {
+			log.Warn("Both --local-addr and --local-interface specified. Ignoring interface.")
```
zakird: It might be better to Fatal here rather than Warn. That will make sure that the user knows of the problem. I'm always bit hesitant to eat incoherent configurations rather than let them pass through with a log entry.

zakird (changes requested):
Overall, this looks great. Thank for you the contribution. I left a couple of minor fixes.

zdns/main.go
stklcode: Most likely one key right of what i intended to type... :see_no_evil:

zdns/main.go
stklcode: Good catch. Moved the info-log behind the loop.

zdns/main.go
stklcode: Changed to fatal and removed the "Ignoring..." part

stklcode:
Added fixes for all three points and rebased the branch onto the current _master_.

zakird:
I've merged this via #203 with an additional commit that uses a fixed version of the backend library. Thank you!!

PR #175, opened 2019-12-29
Do not panic on Layers by definition are suffixes of names

### LINKED ISSUE #162: panic: Layers by definition are suffixes of names
I ran into this message while doing a bunch of DNS lookups. I'm not sure what causes this error, but it would be helpful if zmap just returned an error JSON for the corresponding lookup, rather than dying with a panic.

zakird:
We catch the error and pass back to user instead of panicking. Fixes #162.

### DIFF
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -875,7 +875,7 @@ func (s *Lookup) cachedRetryingLookup(dnsType uint16, dnsClass uint16, name stri
 	if err != nil {
 		s.VerboseLog(depth+2, err)
 		var r Result
-		return r, isCached, zdns.STATUS_AUTHFAIL, nil
+		return r, isCached, zdns.STATUS_AUTHFAIL, err
 	}
 	if name != layer && authName != layer {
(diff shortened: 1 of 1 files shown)


zakird:
Background on what causes this:

Trying lookup the IP for riga.the-wedding.ru
->  A riga.the-wedding.ru -> roots
<- authoritatives for .ru and additionals for those NS. ( :thumbsup: )
->  A riga.the-wedding.ru -> .ru NS
<- authoritatives for the-wedding.ru: ns1.reg.ru, ns2.reg.ru and additionals for those NS ( :thumbsup: )
-> A riaa.the-wedding.ru -> ns2.reg.ru (specifically, 194.58.117.14)
Now I would have hoped for an A record at this point, but instead, you get…
<- ;; AUTHORITY SECTION:
*.the-wedding.ru.	86400	IN	NS	ns1.reg.ru.
*.the-wedding.ru.	86400	IN	NS	ns2.reg.ru.
(which is nonsensical)

dadrian (approved):

PR #176, opened 2020-01-07
Use Cache when rDNS

zakird:
This PR moves up where we perform conversion from IP to `reverse-ip.in-addr.arpa` to be at the very start of the process so that caching works properly. It removes a hack around `in-addr.arpa` in the recursion and replaces it with one that catches the skip layer that when you ask a root for `x.in-addr.arpa`, you never receive an entry for `.arpa`.  The PR also improves the logging around wire requests to indicate the type of request sent to the server.


modules/miekg/miekg.go
```diff
@@ -17,6 +17,79 @@ import (
 	"github.com/zmap/zdns/cachehash"
 )
 
+var typeNames = map[uint16]string{
```
dadrian, line 20: Classic Golang.

modules/miekg/miekg.go
```diff
@@ -1147,6 +1212,14 @@ func (s *Lookup) iterativeLookup(dnsType uint16, dnsClass uint16, name string, n
 }
 
 func (s *Lookup) DoMiekgLookup(name string) (interface{}, []interface{}, zdns.Status, error) {
+	if s.DNSType == dns.TypePTR {
+		var err error
+		name, err = dns.ReverseAddr(name)
+		if err != nil {
+			return nil, nil, zdns.STATUS_ILLEGAL_INPUT, err
+		}
+		name = name[:len(name)-1]
```
dadrian, line 1221: Are you sure you want fall-through here?

dadrian (approved):

modules/miekg/miekg.go
zakird: yeah, I tested it and it gets thrown back to the user correctly. I think it just feels weird because nowhere else do we do pre-processing like this.

PR #177, opened 2020-01-08
include dns resolver in result

zakird:
I think this was an oversight, but while we specify things like what protocol the response was sent over, we don't include the server that sent back the response, which is incredibly useful.

### DIFF
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -118,6 +118,7 @@ type Result struct {
 	Additional  []interface{} `json:"additionals"`
 	Authorities []interface{} `json:"authorities"`
 	Protocol    string        `json:"protocol"`
+	Resolver    string        `json:"resolver"`
 	Flags       DNSFlags      `json:"flags"`
 }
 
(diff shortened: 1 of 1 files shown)


modules/miekg/miekg.go
```diff
@@ -745,6 +746,7 @@ func (s *Lookup) doLookup(dnsType uint16, dnsClass uint16, name string, nameServ
 // Expose the inner logic so other tools can use it
 func DoLookupWorker(udp *dns.Client, tcp *dns.Client, dnsType uint16, dnsClass uint16, name string, nameServer string, recursive bool) (Result, zdns.Status, error) {
 	res := Result{Answers: []interface{}{}, Authorities: []interface{}{}, Additional: []interface{}{}}
+	res.Resolver = nameServer
```
dadrian, line 749: If you're creating the object as a literal on the line above, you should just add this to the literal.

modules/miekg/miekg.go
```diff
@@ -1112,7 +1114,9 @@ func handleStatus(status *zdns.Status, err error) (*zdns.Status, error) {
 	}
 }
 
-func (s *Lookup) iterateOnAuthorities(dnsType uint16, dnsClass uint16, name string, depth int, result Result, layer string, trace []interface{}) (Result, []interface{}, zdns.Status, error) {
+func (s *Lookup) iterateOnAuthorities(dnsType uint16, dnsClass uint16, name string,
+	depth int, result Result, layer string, trace []interface{}) (Result, []interface{}, zdns.Status, error) {
+	//
```
dadrian, line 1119: Remove empty `//` line.

dadrian (approved):
LGTM % nits

PR #181, opened 2020-01-21
performance tune what we cache

zakird:
Currently, when we perform large-scale PTR lookups, our cache fills up with PTR records which doesn't help solve any problems, and in fact slows things down by (1) grabbing a write lock on the cache, and (2) evicting useful data in the cache.

This PR changes our cache before to only cache A/AAAA/NS/DNAME records since those can be used for future lookups. This solution isn't 100% perfect since if we're doing A lookups on a tremendous number of names, those will still end up in the cache. However, that's a _significantly_ more complex problem to address and will be handled separately.

### DIFF
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -671,12 +671,13 @@ func (s *GlobalLookupFactory) AddCachedAnswer(answer interface{}, name string, d
 		// we can't cache this entry because we have no idea what to name it
 		return
 	}
-	// only cache records that can help prevent future iteration.
-	// There's no reason to cache the leaf records that are unlikely to be useful in the future
+	// only cache records that can help prevent future iteration: A(AAA), NS, (C|D)NAME.
+	// This will prevent some entries that will never help future iteration (e.g., PTR)
+	// from causing unnecessary cache evictions.
 	// TODO: this is overly broad right now and will unnecessarily cache some leaf A/AAAA records. However,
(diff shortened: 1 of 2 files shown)


modules/miekg/miekg.go
```diff
@@ -671,6 +671,14 @@ func (s *GlobalLookupFactory) AddCachedAnswer(answer interface{}, name string, d
 		// we can't cache this entry because we have no idea what to name it
 		return
 	}
+	// only cache records that can help prevent future iteration.
+	// There's no reason to cache the leaf records that are unlikely to be useful in the future
+	// TODO: this is overly broad right now and will unnecessarily cache some leaf A/AAAA records. However,
+	// it's a lot of work to understand _why_ we're doing a specific lookup and this will still help
+	// in other cases, e.g., PTR lookups
+	if !(dnsType == dns.TypeA || dnsType == dns.TypeAAAA || dnsType == dns.TypeNS || dnsType == dns.TypeDNAME) {
```
dadrian: Why DNAME and not CNAME?
Why NS?

dadrian (approved):
LGTM, but should cache CName, and you could clarify that leaf does not necessarily mean `A` or `AAAA` in this case.

modules/miekg/miekg.go
dadrian: Oh, there's a `!`

zakird:
fixed comment and added CNAME type

PR #184, opened 2020-01-25
introduce --result-verbosity={short,normal,long,trace}

zakird:
ZDNS output is very verbose today, and most of the data isn't useful for day-to-day operations. This flag allows the user to determine how much information they want returned.


zakird:
<sigh> Tests all fail because ZDNS output is different now. Will fix up tests.

dadrian (approved):

Issue #185, opened 2020-02-02
Add option which passes resolvers using file


wavvs:
Currently if I want to pass 1000 resolvers I need to pass string in -name-servers option, which is not convenient at all. Please, add such an option.

zakird:
We'd be happy to accept a pull request that adds this feature if you can
submit one.

stklcode:
Wondering which implementation would be more convenient. Additional parameter like `--name-servers-file` or @-syntax like `--name-servers @/path/to/file`

zakird:
#187 resolves this issue, so closing.

PR #187, opened 2020-02-02
Add option which passes resolvers using file #185

wavvs:
Tried to resolve issue without introducing new option.


zakird (approved):

zakird:
Looks good. thank you!

PR #189, opened 2020-02-06
sharded cache hash

zakird:
This provides 15-20% at 5K threads and 20-25% at 10K threads performance improvement. While it's tiny bit less fair in eviction, the removal of a global lock over every cache entry more than makes up for it empirically, so I don't see much if any downside. It also allows much higher number of threads without slowdown.

This also changes the RWMutex to a normal mutex since we never used Read vs Write locks anyway.

### DIFF
--- a/cachehash/cachehash.go
+++ b/cachehash/cachehash.go
@@ -24,7 +24,7 @@ type CacheHash struct {
 	l       *list.List
 	len     int
 	maxLen  int
-	mutex   *sync.RWMutex
+	mutex   *sync.Mutex
 	ejectCB func(interface{}, interface{})
 }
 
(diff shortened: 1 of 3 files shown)


cachehash/cachehash.go
```diff
@@ -14,13 +14,17 @@
...
+	"container/list"
+	"sync"
+)
 
 type CacheHash struct {
 	h       map[interface{}]*list.Element
 	l       *list.List
 	len     int
 	maxLen  int
+	mutex   *sync.Mutex
```
dadrian: This should be an anonymous struct, which will automatically add the `Lock()` and `Unlock()` methods to your object.

dadrian: https://talks.golang.org/2012/10things.slide#3

dadrian (approved):
LGTM % nits

PR #192, opened 2020-02-21
add bind version query module

zakird:


modules/bindversion/bindversion.go
```diff
@@ -0,0 +1,79 @@
...
+type Lookup struct {
+	Factory *RoutineLookupFactory
+	miekg.Lookup
+}
+
+func (s *Lookup) DoLookup(_ string) (interface{}, []interface{}, zdns.Status, error) {
+	var res Result
+	innerRes, trace, status, err := s.DoTxtLookup("VERSION.BIND")
+	if status != zdns.STATUS_NOERROR {
+		return res, trace, status, err
```
dadrian, line 39: Do you want to return `res` or `nil` here? I'm not very familiar with the API---are you sure you want an empty result back in the error case?

dadrian (approved):

PR #193, opened 2020-02-22
Name Server Mode

zakird:
This PR introduces the notion of "Name Server Mode", which allows pivoting ZDNS from its normal behavior of:

```cat "list of names" | zdns A --name-servers=[...]```

to 

```cat "list of name servers" | zdns A --name-server-mode --override-name=google.com```

It also supports tuples of `(name,nameServer)`s, e.g.,

```echo "google.com,8.8.8.8" | zdns A```

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ supports mixing and matching both modes by piping in a comma-delimited list of
 ```echo "google.com,8.8.8.8" | ./zdns A``` will send an `A` query for
 `google.com` to `8.8.8.8` regardless of what name servers are specified by
 `--name-servers=` flag. Lines that do not explicitly specify a name server will
-use the servers specified by the OS or `--name-server` flag as would normally
+use the servers specified by the OS or `--name-servers` flag as would normally
 happen.
 
 
--- a/README.md
+++ b/README.md
@@ -153,6 +153,30 @@ flag and specifying a list of fields, e.g., `--include-fields=flags,resolver`.
 Additional fields are: class, protocol, ttl, resolver, flags.
 
 
+Name Server Mode
+----------------
+
+By default ZDNS expects to receive a list of names to lookup on a small number
+of name servers. For example:
+
+```echo "google.com" | ./zdns A --name-servers=8.8.8.8,8.8.4.4```
+
+However, there are times where you instead want to lookup the same name across
+a large number of servers. This can be accomplished using _name server mode_.
+For example:
+
+```echo "8.8.8.8" | ./zdns A --override-name="google.com"```
+
+Here, every line piped in ZDNS is sent an A query for `google.com`. ZDNS also
+supports mixing and matching both modes by piping in a comma-delimited list of
+`name,nameServer`. For example:
+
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -153,6 +153,30 @@ flag and specifying a list of fields, e.g., `--include-fields=flags,resolver`.
...
+By default ZDNS expects to receive a list of names to lookup on a small number
+of name servers. For example:
+
+```echo "google.com" | ./zdns A --name-servers=8.8.8.8,8.8.4.4```
+
+However, there are times where you instead want to lookup the same name across
+a large number of servers. This can be accomplished using _name server mode_.
+For example:
+
+```echo "8.8.8.8" | ./zdns A --name-server-mode= --override-name="google.com"```
```
dadrian:
```suggestion
```echo "8.8.8.8" | ./zdns A --name-server-mode --override-name="google.com"```
```

README.md
```diff
@@ -153,6 +153,30 @@ flag and specifying a list of fields, e.g., `--include-fields=flags,resolver`.
...
+
+Here, every line piped in ZDNS is sent an A query for `google.com`. ZDNS also
+supports mixing and matching both modes by piping in a comma-delimited list of
+`name,nameServer`. For example:
+
+```echo "google.com,8.8.8.8" | ./zdns A``` will send an `A` query for
+`google.com` to `8.8.8.8` regardless of what name servers are specified by
+`--name-servers=` flag. Lines that do not explicitly specify a name server will
+use the servers specified by the OS or `--name-server` flag as would normally
```
dadrian:
```suggestion
use the servers specified by the OS or `--name-servers` flag as would normally
```

lookup.go
```diff
@@ -99,21 +111,25 @@ func doLookup(g *GlobalLookupFactory, gc *GlobalConf, input <-chan interface{},
...
 				rawName, rank = parseAlexa(line)
 				res.AlexaRank = rank
+			} else if gc.NameServerMode {
+				nameServer = AddDefaultPortToDNSServerName(line)
 			} else {
-				rawName = line
+				rawName, nameServer = parseNormalInputLine(line)
 			}
-			lookupName, changed := makeName(rawName, gc.NamePrefix)
+			lookupName, changed = makeName(rawName, gc.NamePrefix, gc.NameOverride)
```
dadrian, line 126: The interaction between `makeName` and `NameServerMode` is non-obvious, but short of splitting up each style of parsing into it's own function and doing one or the other, I'm not sure this gets any simpler.

integration_tests.py
```diff
@@ -37,6 +39,17 @@ def run_zdns(self, command, name):
         o = subprocess.check_output(c, shell=True)
         return c, json.loads(o.rstrip())
 
+    def run_zdns_multiline(self, command, names):
```
dadrian, line 42: It doesn't look like you have a test that calls this?

dadrian (approved):
Hard for me to tell if you missed passing a nameserver somewhere, but your changes look reasonable.

PR #195, opened 2020-02-23
Round out DNS Type Support for ZDNS

zakird:
This fixes some awfully repeated code and adds native support for all types that miekg has parsers for.

PR #196, opened 2020-02-24
WIP: Individual record locks

zakird:
This is based off of #195, so I'll deal with testing once that's merged. It needs testing to see if it actually improves performance.


zakird:
This does not do good things.

PR #199, opened 2020-02-27
Name server mode fix

zakird:


stklcode (commented):
Fixed the name server mode issue with BINDVERSION without breaking behavior for other modules.

```bash
$ echo example.com | zdns A -name-server-mode
FATA[0000] Static Name must be defined with --override-name in --name-server-mode unless DNS module does not expect names (e.g., BINDVERSION).

$ echo dns.example.com | zdns BINDVERSION -name-server-mode
{"data":{"version":"Example DNS Server v1.0"},"status":"NOERROR","timestamp":"2020-02-27T19:10:17+01:00"}
```
Correct result on first sight.

----

However the BINDVERSION module is not yet perfect... (should not block this PR imho)
```bash
$ echo dns.example.com | zdns BINDVERSION -name-server-mode -result-verbosity long
{"class":"IN","data":{"version":"Example DNS Server v1.0"},"status":"NOERROR","timestamp":"2020-02-27T19:17:32+01:00"}
```

I'm missing the name server here. The above result seems useless for any deeper analysis of batch requests.
Also the reported class is incorrect, as the requested record definitely lives in _CH_ not _IN_. As "version" is already abstracted, we should simply omit the class even in _long_ mode.

Reference record using the old fashioned way:
```bash
$ echo version.bind | zdns TXT -class CHAOS -name-servers dns.example.com -result-verbosity long
{"class":"CH","data":{"answers":[{"answer":"Example DNS Server v1.0","class":"CH","name":"version.bind","ttl":0,"type":"TXT"}],"authorities":[{"answer":"version.bind.","class":"CH","name":"version.bind","ttl":0,"type":"NS"}],"flags":{"authenticated":false,"authoritative":true,"checking_disabled":false,"error_code":0,"opcode":0,"recursion_available":false,"recursion_desired":true,"response":true,"truncated":false},"protocol":"udp","resolver":"dns.example.com:53"},"name":"version.bind","status":"NOERROR","timestamp":"2020-02-27T19:22:13+01:00"}

zakird:
100% agree with all of this. Going to move into a different issue, very different part of code base affected and doesn't need to block this particular PR.

dadrian (approved):

Issue #209, opened 2020-04-14
Generalize Input/Output handlers to io.Reader, io.Writer


jcrowgey:
I'm using zdns in a larger context (importing it) and it's inconvenient to require file paths when in-memory byte-streams will do.  Looking at the exsting iohandlers/ (ie, the implementations of InputHandler and OutputHandler interfaces), there's only the "file" version, which requires a path on disk (or "-" as a stand-in).   After opening the file, these handlers generally uses library functions that are abstracted to require io.Reader/io.Writer anyway (eg, calls to bufio.NewScanner, or WriteString).

To solve my issue (which is that my data isn't necessarily in a file on disk), I've got a couple of paths forward: I can create my own Input/Output handlers for "stream" which will basically just require the io.Reader io.Writer types to be passed to them.  If I go that route and that would be useful for others, I'm happy to offer a PR with those implementations change.  Alternatively, I can modify the existing "file" handlers so that they're generalized to io.Reader, io.Writer and modify znds/main.go to do the open/close based on what's been asked for by the user.

Effectively, it seems like the library/package aspects of zdns shouldn't really care whether the input/output is a file or not, and it's not clear to me that we need separate implementations of the InputHandler/OutputHandler interfaces in order to support library users vs command line users.  That's because the command-line users are just the same as the library users after a call to os.Open.  And the command-line app can handle that call before init'ing the factory.

Any thoughts or suggestions?

dadrian:
We'd be happy to take a PR that exposes stream-based handlers and defers the open/close to main.

PR #210, opened 2020-04-14
File I/O Handlers implement interface

jcrowgey:
We already have an interface for InputHandler and OutputHandler.  This
change removes the registration system and instead relies on the
implementations of the interface methods to do the right thing.  The
GlobalConf object now holds a pointer to an implementation the
interfaces.

    * CLI users could only use the "file" handler type anyway, so this
    option is removed.
    * The registration system (and the associated getters) are removed,
    Since the config holds an implementation, the factory can check that
    the config's InputHandler and OutputHandler are not nil before
    calling their  methods.
    * Some minor changes to Golang's style: for example, avoiding "else"
    after a Fatal keeps the happy-path less indented

This change leaves the functionality of the CLI unmodified (I hope!)
while also making it easier for programs using zdns as a library to
create their own implementations of the I/OHandler interfaces---they can
simply stick them into a znds.GlobalConfig object instead of having to
go through registration.

### DIFF
--- a/conf.go
+++ b/conf.go
@@ -45,8 +45,8 @@ type GlobalConf struct {
 	LocalAddrSpecified   bool
 	LocalAddrs           []net.IP
 
-	InputHandler  string
-	OutputHandler string
+	InputHandler  InputHandler
+	OutputHandler OutputHandler
 
 	InputFilePath    string
 	OutputFilePath   string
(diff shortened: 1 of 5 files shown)


jcrowgey:
This is related to #209.  It doesn't actually create the stream handlers, but it does make the I/O Handler interface more easily usable from external packages.

zakird (approved):
This looks reasonable to me. Asking @dadrian to take a look as well.

zakird:
This was superseded by #211, so closing.

PR #211, opened 2020-04-15
implement streamhandlers, simplifies I/OHandler interface

### LINKED ISSUE #209: Generalize Input/Output handlers to io.Reader, io.Writer

jcrowgey:
This change adds streamhandlers which can be use by programs which import
zdns who need other types of input/output handlers beyond the pipes and
disk-files provided by the filehandlers.

* "New" constructors are provided for the File and Stream handler
implementations.  This allows them to keep their filepath (or
reader/writer) private, while also allowing them to be created by
packages which import the handlers.

* The {Input,Output}Handler interfaces are simplified, the Inititalize
method is removed (as are the calls to it in lookup).  This method
wasn't being used by the only existing implementation beyond as a way to
pass the filename to the Handler implementation.  The "New" constructors
fulfill this requirement in a way which is (hopefully) more transparent
and easier to use.

* The "file" package is subsumed by the "iohandlers" package, which
provides the two implementations.

* zdns/main.go is modified to use the File{In,Out}putHandlers

Fixes #209


jcrowgey:
Here's another approach at #209.  This one actually provides the streamhandlers alongside the filehandlers.  This approach is built on the PR in #210, so you can consider it a follow-on.

jcrowgey:
One further change, which I considered but didn't implement here, would be to have the File{In,Out}putHandler embed the Stream{In,Out}putHandler and then simply try to open the filepath on some call to New.

zakird (approved):

Issue #212, opened 2020-04-16
Should we use upstream's TypeToStream?


jcrowgey:
miekg/dns has a complete (I believe) TypeToString map here [1].  We have a partial map (just the answers, I suppose (ie, excluding things like AXFR and ANY)), here [2].  Grepping zdns suggest that's only actually used for logging though:

```

$ grep "typeNames" ./* -r
./modules/miekg/answers.go:var typeNames = map[uint16]string{
./modules/miekg/miekg.go:	s.VerboseLog(1, "****WIRE LOOKUP*** ", typeNames[dnsType], " ", name, " ", nameServer)
Binary file ./zdns/zdns matches
```
It seems, then, like we don't actually need to maintain the local map.

[1]: https://github.com/miekg/dns/blob/67373879ce327b5fd112d9301d0a4d62bad6b904/ztypes.go#L86
[2]: https://github.com/zmap/zdns/blob/2d530641f50803270bae0aafe591ff5c2ee7d458/modules/miekg/answers.go#L12

zakird:
addressed in 213

PR #213, opened 2020-04-16
Remove local typeNames map, use dns.TypeToString

jcrowgey:
This removes the partial map and relies on the one from miekg's dns
package.


jcrowgey:
Fixes #212

zakird (approved):

zakird:
Thank you!

PR #216, opened 2020-04-23
Don't repeat back-to-back types in parameter lists

jcrowgey:
This makes parameter lists a bit easier to read and conforms with
typical Golang style to use type elisions where applicable.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -47,7 +47,7 @@ import (
 // one Lookup per IP/name/connection ==========================================
 //
 type Lookup interface {
-	DoLookup(name string, nameServer string) (interface{}, []interface{}, Status, error)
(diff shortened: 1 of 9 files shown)


zakird (changes requested):
All for the types checks, but looks like you accidentally committed requirements into go.sum and go.mod that aren't needed be the committed codebase.

jcrowgey:
> All for the types checks, but looks like you accidentally committed requirements into go.sum and go.mod that aren't needed be the committed codebase.

Heh, whoops.  Let me clean that up.  Sorry.

jcrowgey:
Fixed, squashed and force-pushed for a single commit merge (hopefully) :)

zakird (approved):

Issue #220, opened 2020-05-22
Set DO flag when doing NSEC lookups.


simonvik:
Some DNS servers wont answer on NSEC querys unless the DO flag has been set.

Easiest patch is to add `	m.SetEdns0(1232, true)` in `DoLookupWorker` but i dont know if its desired to do this on all querys. 

Perhaps some flag is needed for EDNS buffer size and DO?

zakird:
I think that’s probably fine to set globally for now. Can you submit PR?

zakird:
Addressed by #323

PR #221, opened 2020-05-22
Set EDNS Do flag, Buf size. Add NextDomain, TypeBitMap for NSEC3

simonvik:
This patch sets the UDP Buffer size, this might lead to more TC answers but it also follows the new Flag Day recommendations. 

Solves #220


zakird (approved):

simonvik:
I see that one check failes now due to the "TC" test being less than 1232, dont know how that should be solved, i can lower the bufsize perhaps?

simonvik:
The test seems to fail due to the TC testing being smaller than 1232, should i lower the limit or should the test get more strings?

zakird:
I just saw that too. This should have automatically transitioned over to TCP rather than stay with UDP and end up truncated, unless it was marked `--udp-only`. So somewhere, it does seem like there's a bug.

simonvik:
The test runs as "./zdns/zdns TXT --udp-only" and expect TC but wont get it now with 1232 bufsize.

zakird:
Oh, I see. There seem to be two issues here, actually.

The first, we can just increase the size of the `TXT` record to make it large enough to be truncated. The other problem seems to be that the result comes back as `TCP` even though it's not (which threw me off in that earlier comment).

simonvik:
the TLSA test also fail when DO is set, trying to figure out why.

simonvik:
_25._tcp.mail.ietf.org is DNSSEC signed, the rest is not. The test framework doesn't handle receiving RRSIGs very well.

spencerdrak:
Looks like it's a been a bit since the last activity here - if @simonvik still wants to address this then great, otherwise, I'm happy to take a look at it.

simonvik:
Feel free to take over! Think the test was fighting me last time :)

phillip-stephens:
Looks like this was completed in  #323. I'll close this PR!

PR #226, opened 2020-06-05
Actually fix incorrect extract authorities logic

paul-pearce:
PR #225 was buggy. It left in a return statement it shouldn't have.
As a result of fixing this bug, it revealed incorrect behavior
(repeated retries) on NXDOMAINS which this commit also fixes. Lastly
it corrects the debug msgs which were also not accurate.

### DIFF
--- a/modules/miekg/miekg.go
+++ b/modules/miekg/miekg.go
@@ -615,7 +615,7 @@ func (s *Lookup) checkGlue(server string, depth int, result Result) (Result, zdn
 		if !ok {
 			continue
 		}
-		if (ans.Type == "A" || ans.Type == "AAAA") && strings.TrimSuffix(ans.Name, ".") == server {
+		if ans.Type == "A" && strings.TrimSuffix(ans.Name, ".") == server {
 			var retv Result
 			retv.Authorities = make([]interface{}, 0)
 			retv.Answers = make([]interface{}, 0)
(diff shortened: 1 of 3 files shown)


modules/miekg/miekg.go
```diff
@@ -752,13 +752,20 @@ func (s *Lookup) iterateOnAuthorities(dnsType, dnsClass uint16, name string,
 
 			} else {
 				// otherwise we hit a status we know
-				return r, trace, *new_status, err
+
+				// If we get here, querying this domain all the way down yielded an NXDOMAIN. Thats a fatal error
+				//  don't try other auths
+				if status == zdns.STATUS_NXDOMAIN {
```
zakird: Couldn't we just move this to L746?

paul-pearce: No, because of the idiom of returning new_status not status. So we need to place this after the handle_status call and check.

I dislike the structure of handle_status, but alas, that's a different PR.

modules/miekg/miekg.go
```diff
@@ -82,6 +82,13 @@ func TranslateMiekgErrorCode(err int) zdns.Status {
 	return zdns.Status(dns.RcodeToString[err])
 }
 
+func isStatusAnswer(s zdns.Status) bool {
```
dadrian, line 85: This is a horrible name for a function

dadrian (approved):

PR #241, opened 2021-01-06
Allow to pass metadata from input to output similar to Alexa rank

janrueth:
This PR adds a new option (`-metadata` / `GlobalConf.MetadataFormat`) that allows to pass additional metadata from the input to the output for each name that is to be resolved. This is very helpful in situations where a single zdns process is fed from multiple sources and one would e.g., like to keep a reference to said source.

When setting `MetadataFormat` to true, the input is expected to be in the form of:
 name,METADATA

So similar to the Alexa format but a normal string is allowed but the order is inverted.
Inverting the format has the advantage that further comma or even a full json line can be easily embedded while keeping the implementation simple.

A new key (`meta_data`) is added to the output containing the data.

Example:

 ```
echo 'www.google.com,{"source": ".com.tld", "reason": "just a test"}' | ./zdns A -metadata -name-servers="8.8.8.8"
{"data":{"answers":[{"answer":"172.217.21.196","class":"IN","name":"www.google.com","ttl":161,"type":"A"}],"protocol":"udp","resolver":"8.8.8.8:53"},"meta_data":"{\"source\": \".com.tld\", \"reason\": \"just a test\"}","name":"www.google.com","status":"NOERROR","timestamp":"2021-01-06T13:59:19+01:00"}
```

### DIFF
--- a/conf.go
+++ b/conf.go
@@ -25,6 +25,7 @@ type GlobalConf struct {
 	IterationTimeout      time.Duration
 	Retries               int
 	AlexaFormat           bool
+	MetadataFormat        bool
 	NameServerInputFormat bool
 	IterativeResolution   bool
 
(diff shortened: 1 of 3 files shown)


zakird:
Overall this looks good. We might want to add a second flag to indicate whether the metadata is JSON so that we don't escape all of it out. It seems likely that folks will pass in JSON here.

janrueth:
I also thought about that option, however in this case I would rather treat the whole input as json and declare a field or path to the domain name and either merge this json with the result or add it as a subobject.

zakird:
That makes sense. That's what we do in https://github.com/zmap/zannotate -- we just inject the annotation data as a sub-record.

zakird (approved):

Issue #253, opened 2021-10-01
SPF prefix should be case insensitive


bayerhonza:
[RFC 7208](https://datatracker.ietf.org/doc/html/rfc7208) notes that literal text strings in ABNF are case-insensitive.

This means that following SPF records are valid but do not apear as valid records while using zdns with SPF module:

| SPF record | zdns result|
| -------- | ----------------|
| v=spf1 | NOERROR |
| V=sPf1 | **NORECORD** |
| v=SPF1 | **NORECORD** |

Checking the SPF prefix ("v=spf1") should be case-insensitive in order to conform to RFC.

zakird:
Thanks! Would you be willing to submit a PR to address this?

bayerhonza:
Yes, I will submit a PR together with #252

zakird:
Fantastic, thank you!!

PR #254, opened 2021-10-01
fixes SPF and DMARC prefix checks (#252 #253)

### LINKED ISSUE #252: Prefix in DMARC module differs from RFC
Hello,

while using the DMARC module, we found that its implementation [modules/dmarc/dmarc.go#L56](https://github.com/zmap/zdns/blob/ed5b6828b56753f42903a3b7b82ade161e65ec02/modules/dmarc/dmarc.go#L56) checks only the first 7 characters ("v=DMARC") of DMARC TXT record and not the ABNF defined in [RFC 7489](https://datatracker.ietf.org/doc/html/rfc7489) as "v=DMARC1" with optional whitespaces.

This can lead to cases when a DMARC of a domain is not set up correctly (e.g. "v=DMARCv1") and zdns' result status is "NOERROR" even though the DMARC configuration is not valid.

Is it an intentional behviour?

Thanks.

bayerhonza:
resolves #252 #253 

Added prefix checks for TXT records (DMARC and SPF) based on regular expressions

DMARC prefix is now checked correctly as well as case-insensitive SPF prefix


modules/dmarc/dmarc.go
```diff
@@ -18,8 +18,11 @@ import (
 	"github.com/miekg/dns"
 	"github.com/zmap/zdns"
 	"github.com/zmap/zdns/modules/miekg"
+	"regexp"
 )
 
+const dmarcPrefixRegexp = "v[\x09\x20]*=[\x09\x20]*DMARC1[\x09\x20]*;[\x09\x20]*"
```
dadrian: Are \x09 and \x20 whitespace characters? Should we instead be using `[\s]*`?

dadrian (commented):
Sorry for the delay on the review.

spencerdrak:
This also looks fine to me pending the resolution of @dadrian's comment above, assuming the only acceptable whitespace that might be encountered is tab (`\x09`) and space (`\x20`). `\s` will also catch newlines as whitespace, so it may be more useful to only catch tabs and spaces.

modules/dmarc/dmarc.go
bayerhonza: I used the ABNF from the [DMARC RFC](https://datatracker.ietf.org/doc/html/rfc7489) that states:

`dmarc-version   = "v" *WSP "=" *WSP %x44 %x4d %x41 %x52 %x43 %x31`

WSP comes from [ABNF RFC](https://datatracker.ietf.org/doc/html/rfc5234) and is defined as:

` WSP            =  SP / HTAB
                                ; white space`

`[\s]*` is not equal to `[\x09\x20]*` as the first one accepts also newline, form feed, and carriage return. [[1]](https://pkg.go.dev/regexp/syntax)

However, I now see that `"v"` should actually be case insensitive as ABNF strings are case insensitive [[ABNF 2.3]](https://datatracker.ietf.org/doc/html/rfc5234). So the code should rather look like this:

`const dmarcPrefixRegexp = "[vV][\x09\x20]*=[\x09\x20]*DMARC1[\x09\x20]*;[\x09\x20]*"`

What do you think?

spencerdrak: I agree with that - @dadrian you have any thoughts?

zakird: Hard to argue against matching the RFC! Let's update to case insensitive and ship it!

bayerhonza:
I updated the code as discussed: "V" in DMARC prefix is now case-insensitive.

spencerdrak:
@bayerhonza we merged in some other work and looks like we have some conflicts now. If you don't mind taking a look at that, we can merge once the conflicts are resolved. Otherwise, I'll pull a copy and sort out the conflicts when I've got the chance. 

Thanks! Sorry for the delay.

bayerhonza:
I resolved the conflicts. I think we can either close this PR or #265 as there are duplicate. I only created #265 because I added a testing library.

dadrian (approved):

Issue #125, opened 2017-09-06
Windows not supported


ghost:
The config line in main.go doesn't support the Windows operating system. I had to change the line to the following with a custom text file to register nameservers:

`config_file := flags.String("conf-file", "c:\\Users\\Adhir\\Documents\\resolv.conf", "config file for DNS servers")`

Ideally it should determine what OS you're on then select the file accordingly if possible

matricali:
Perhaps it is convenient to incorporate a .conf file with default values ​​inside the repository, and load it when the platform is Windows instead of "/etc/hosts" :thinking: 

What do you think?

zakird:
I think that's fine, though if there's a conf file, we'd want it to be able to handle all the options passed in on the command line. I don't think it makes sense to have a conf that only specifies the equivalent of `/etc/hosts`. None of the core maintainers of ZDNS use Windows, so this is something that would need an external contribution.

matricali:
I do not use Windows as my main platform, but I can help by sending a pull request I think.

I'm thinking something like this:

```
config_file_path = "/etc/resolv.conf"

IF PLATFORM == Windows THEN
    config_file_path = HOME_DIR + "\\resolv.conf.zmap"
    IF path NOT EXISTS THEN
        WRITE DEFAULT VALUES to HOME_DIR + "\\resolv.conf.zmap"
    ENDIF
ENDIF
```
The default content for `resolv.conf.zmap` would be something like this:
```
nameserver 8.8.8.8
nameserver 8.8.4.4
```
Perhaps it would be appropriate to print a warning that the file was created with its default values.

zakird:
I'm not sure if this really needs an extra conf file. We could just default to set a set of resolvers if nothing is passed in on the command line and we can't find an equivalent to `/etc/resolv.conf`. 

There's also probably a way to just ask Windows what its resolvers are, similar to reading `/etc/resolv.conf`.

matricali:
> I do not use Windows as my main platform, but I can help by sending a pull request I think.
> 
> I'm thinking something like this:
> 
> ```
> config_file_path = "/etc/resolv.conf"
> 
> IF PLATFORM == Windows THEN
>     config_file_path = HOME_DIR + "\\resolv.conf.zmap"
>     IF path NOT EXISTS THEN
>         WRITE DEFAULT VALUES to HOME_DIR + "\\resolv.conf.zmap"
>     ENDIF
> ENDIF
> ```
> 
[...]

PR #257, opened 2021-11-19
- fixed startup error on windows

ksv87:
- added verification of the existence of the configuration file, if the file does not exist, DNS servers 8.8.8.8 and 8.8.4.4 are used, no additional configuration files are generated
- https://github.com/zmap/zdns/issues/125

### DIFF
--- a/lookup.go
+++ b/lookup.go
@@ -16,7 +16,6 @@ package zdns
 
 import (
 	"encoding/json"
-	"errors"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


lookup.go
```diff
@@ -34,17 +35,21 @@ type routineMetadata struct {
...
 func GetDNSServers(path string) ([]string, error) {
-	c, err := dns.ClientConfigFromFile(path)
-	if err != nil {
-		return []string{}, err
-	}
 	var servers []string
-	for _, s := range c.Servers {
-		if s[0:1] != "[" && strings.Contains(s, ":") {
-			s = "[" + s + "]"
+	if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
```
zakird: I don't think that it makes sense for defaults to be specified deep in a function like this. Rather, they should be specified as default values elsewhere, e.g., https://github.com/zmap/zdns/blob/master/zdns/main.go. It also likely makes more sense for them to be added in if the call to `GetDNSServers` fails, rather than being specified in the call itself. Can you investigate trying to move this check upwards in the ZDNS stack?

zakird (changes requested):

lookup.go
ksv87: Ok

ksv87: done

ksv87: @zakird Did I do wrong?

zakird (approved):
Looks good. Added a quick warning and declared as const.

ksv87:
image
travis-ci.org it no longer works, and you need a check

PR #259, opened 2021-12-12
Migrate Travis CI workflow to GH actions

stklcode:
The Travis CI workflow is currently broken, because migration to travis-ci.com is required.
This PR proposes a switch to GH actions as new primary CI environment. Go version and OS unchanged for now.

Proof of concept run: https://github.com/stklcode/zdns/runs/4497939698

There are some implications:
* Default Python environment is running Python 3.8, so some functions, esp. sorting a list of dictionaries are broken. Added some updates to the test script to explicitly sort by "name".
* The latest Ubuntu 20.04 containers utilize _systemd-resolved_ and ZDNS refuses to cooperate with the 127.0.0.53 stub resolver from _/etc/resolv.conf_. (same for my local Fedora dev environment, nothing new) Overriding this for integrations tests makes the default resolver transparent without explicitly linking the configuration file in the test script.
* The vantage point is different. Travis builds run on GCE and AWS machines while hosted GH runners live in Azure. The default DNS resolver is 168.63.129.16 which works fine except for TCP requests.
  I've added Google DNS for the large TXT TCP tests which are the only two cases affected by this issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: CI
+on: [ push, pull_request ]
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.13'
+      - name: Build
+        run: |
+          go version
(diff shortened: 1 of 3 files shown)


zakird (approved):
Thank you for doing this! This has been on my TODO for a while now and I haven't had a chance to do.

PR #262, opened 2022-01-27
Add checks for IPv4 and IPv6

vishalmohanty:
This diff does the following

- Adds checks to make sure `A` returns `IPv4` and `AAAA` returns `IPv6`.
- Adds unit tests for the IP checks
- Adds some additional unit tests for loops, empty non-terminal, non-existent domains

### DIFF
--- a/modules/alookup/alookup.go
+++ b/modules/alookup/alookup.go
@@ -17,6 +17,7 @@ package alookup
 import (
 	"errors"
 	"flag"
+	"net"
 	"strings"
 
 	"github.com/zmap/dns"
@@ -43,13 +44,47 @@ func (s *Lookup) DoLookup(name, nameServer string) (interface{}, zdns.Trace, zdn
 	return s.DoTargetedLookup(name, nameServer)
 }
 
+// Verify that A record is indeed IPv4 and AAAA is IPv6
+func verifyAddress(ansType string, ip string) bool {
+	isIpv4 := net.ParseIP(ip).To4() != nil
+	isIpv6 := net.ParseIP(ip).To16() != nil && strings.Contains(ip, ":")
+	if ansType == "A" {
+		return isIpv4
+	} else if ansType == "AAAA" {
+		return isIpv6
+	}
+	return !isIpv4 && !isIpv6
+}
+
+func populateResults(records []interface{}, dnsType uint16, candidateSet map[string][]miekg.Answer, cnameSet map[string][]miekg.Answer, garbage map[string][]miekg.Answer) {
+	for _, a := range records {
+		// filter only valid answers of requested type or CNAME (#163)
+		if ans, ok := a.(miekg.Answer); ok {
+			lowerCaseName := strings.ToLower(ans.Name)
+			// Verify that the answer type matches requested type
+			if verifyAddress(ans.Type, ans.Answer) {
+				ansType := dns.StringToType[ans.Type]
+				if dnsType == ansType {
+					candidateSet[lowerCaseName] = append(candidateSet[lowerCaseName], ans)
+				} else if ok && dns.TypeCNAME == ansType {
+					cnameSet[lowerCaseName] = append(cnameSet[lowerCaseName], ans)
+				} else {
+					garbage[lowerCaseName] = append(garbage[lowerCaseName], ans)
+				}
+			} else {
+				garbage[lowerCaseName] = append(garbage[lowerCaseName], ans)
+			}
+		}
+	}
+}
+
 func (s *Lookup) doLookupProtocol(name, nameServer string, dnsType uint16, candidateSet map[string][]miekg.Answer, cnameSet map[string][]miekg.Answer, origName string, depth int) ([]string, []interface{}, zdns.Status, error) {
 	// avoid infinite loops
 	if name == origName && depth != 0 {
-		return nil, make([]interface{}, 0), zdns.STATUS_ERROR, errors.New("Infinite redirection loop")
+		return nil, make([]interface{}, 0), zdns.STATUS_ERROR, errors.New("infinite redirection loop")
 	}
 	if depth > 10 {
-		return nil, make([]interface{}, 0), zdns.STATUS_ERROR, errors.New("Max recursion depth reached")
+		return nil, make([]interface{}, 0), zdns.STATUS_ERROR, errors.New("max recursion depth reached")
 	}
 	// check if the record is already in our cache. if not, perform normal A lookup and
 	// see what comes back. Then iterate over results and if needed, perform further lookups
@@ -64,34 +99,8 @@ func (s *Lookup) doLookupProtocol(name, nameServer string, dnsType uint16, candi
 			return nil, trace, status, err
 		}
 
-		for _, a := range miekgResult.(miekg.Result).Answers {
-			// filter only valid answers of requested type or CNAME (#163)
-			if ans, ok := a.(miekg.Answer); ok {
-				lowerCaseName := strings.ToLower(ans.Name)
-				ansType := dns.StringToType[ans.Type]
-				if dnsType == ansType {
-					candidateSet[lowerCaseName] = append(candidateSet[lowerCaseName], ans)
-				} else if ok && dns.TypeCNAME == ansType {
-					cnameSet[lowerCaseName] = append(cnameSet[lowerCaseName], ans)
-				} else {
-					garbage[lowerCaseName] = append(garbage[lowerCaseName], ans)
-				}
-			}
-		}
-		for _, a := range miekgResult.(miekg.Result).Additional {
-			// filter only valid answers of requested type or CNAME (#163)
-			if ans, ok := a.(miekg.Answer); ok {
-				lowerCaseName := strings.ToLower(ans.Name)
-				ansType := dns.StringToType[ans.Type]
-				if dnsType == ansType {
-					candidateSet[lowerCaseName] = append(candidateSet[lowerCaseName], ans)
-				} else if ok && dns.TypeCNAME == ansType {
-					cnameSet[lowerCaseName] = append(cnameSet[lowerCaseName], ans)
-				} else {
-					garbage[lowerCaseName] = append(garbage[lowerCaseName], ans)
-				}
-			}
-		}
+		populateResults(miekgResult.(miekg.Result).Answers, dnsType, candidateSet, cnameSet, garbage)
+		populateResults(miekgResult.(miekg.Result).Additional, dnsType, candidateSet, cnameSet, garbage)
 	}
 	// our cache should now have any data that exists about the current name
 	if res, ok := candidateSet[name]; ok && len(res) > 0 {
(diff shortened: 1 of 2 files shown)


zakird:
@spencerdrak can you take a first pass review?
@vishalmohanty can you please fix conflicts?

modules/alookup/alookup.go
```diff
@@ -43,6 +44,18 @@ func (s *Lookup) DoLookup(name, nameServer string) (interface{}, zdns.Trace, zdn
 	return s.DoTargetedLookup(name, nameServer)
 }
 
+// Verify that A record is indeed IPv4 and AAAA is IPv6
+func verifyAddress(ansType string, ip string) bool {
+	isIpv4 := net.ParseIP(ip).To4() != nil
+	isIpv6 := net.ParseIP(ip).To16() != nil
```
codyprime: `To16()` doesn't make any guarantees the address is IPv6; it will return non-nil for both IPv4 and IPv6, it merely converts either IPv4 or IPv6 into 16-byte representation:

https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/net/ip.go;l=226

if `net.ParseIP(ip).To16()` returns non-nil, you can then check for `':'` in `ip` (the string) to verify it is IPv6

modules/alookup/alookup.go
```diff
@@ -68,11 +81,16 @@ func (s *Lookup) doLookupProtocol(name, nameServer string, dnsType uint16, candi
 			// filter only valid answers of requested type or CNAME (#163)
 			if ans, ok := a.(miekg.Answer); ok {
 				lowerCaseName := strings.ToLower(ans.Name)
-				ansType := dns.StringToType[ans.Type]
-				if dnsType == ansType {
-					candidateSet[lowerCaseName] = append(candidateSet[lowerCaseName], ans)
-				} else if ok && dns.TypeCNAME == ansType {
-					cnameSet[lowerCaseName] = append(cnameSet[lowerCaseName], ans)
+				// Verify that the answer type matches requested type
```
spencerdrak: Only thing I might bring up is that we've got some duplication going on here with what happens in each loop, it may be cleaner to pull that logic out. Thoughts?

spencerdrak (commented):
Overall looks good except for the two comments that are here, lmk what you think @vishalmohanty!

modules/alookup/alookup.go
gakiwate: Is the answer always an A/AAAA/CNAME record? What if you query for a NS record? Does that go through a different logic?

modules/alookup/alookup_test.go
```diff
@@ -263,6 +263,94 @@ func TestDoLookup(t *testing.T) {
...
+	}
+
+	res, _, status, _ = l.DoLookup("example.com", "")
+	if status != zdns.STATUS_NO_ANSWER {
+		t.Errorf("Expected NO_ANSWER status, got %v", status)
+	} else if res != nil {
+		t.Errorf("Expected no results, got %v", res)
+	}
+
+	// Case 12: Loops in CNAME should terminate with no results
```
gakiwate: I wonder if this should terminate with no results or throw an error? For example, if we want to find how many CNAMEs chains have loops in them? Then this will hide those errors from us no? @spencerdrak any thoughts?

[Relevant RFC 1034](https://www.rfc-editor.org/rfc/rfc1034.html#section-3.6.2)
> CNAME chains should be followed and CNAME loops signalled as an error

modules/alookup/alookup_test.go
```diff
@@ -263,6 +263,94 @@ func TestDoLookup(t *testing.T) {
...
+	res, _, status, _ = l.DoLookup("intermediate.example.com", "")
+	if status != zdns.STATUS_NO_ANSWER {
+		t.Errorf("Expected STATUS_NO_ANSWER status, got %v", status)
+	} else if res != nil {
+		t.Errorf("Expected no results, got %v", res)
+	}
+
+	// Case 14: Non-existent domain
+	res, _, status, _ = l.DoLookup("nonexistent.example.com", "")
+	if status != zdns.STATUS_NO_ANSWER {
```
gakiwate: The error here should be NXDOMAIN no?

modules/alookup/alookup_test.go
gakiwate: I would recommend checking how Google and Cloudflare deal with the loops and see if we are consistent with them?

modules/alookup/alookup_test.go
```diff
@@ -263,6 +263,94 @@ func TestDoLookup(t *testing.T) {
...
+	}
+
+	res, _, status, _ = l.DoLookup("example.com", "")
+	if status != zdns.STATUS_NO_ANSWER {
+		t.Errorf("Expected NO_ANSWER status, got %v", status)
+	} else if res != nil {
+		t.Errorf("Expected no results, got %v", res)
+	}
+
+	// Case 13: Empty non-terminal domain should give no answer
```
gakiwate: This should be no answer but also no error? Does STATUS_NO_ANSWER imply NOERROR?

gakiwate (commented):
At a high level, my feedback is to make sure that the RCODEs (NXDOMAIN, NOERROR) that your test suite returns is consistent with how other resolvers do it/RFCs recommend to do it.

modules/alookup/alookup.go
vishalmohanty: Done. Thanks!

modules/alookup/alookup.go
vishalmohanty: @spencerdrak moved it to `populateResults`.

@gakiwate `verifyAddress` only checks for A record containing IPv4 and AAAA containing IPv6. For any other record, we return true

modules/alookup/alookup_test.go
vishalmohanty: Yes. Updated

modules/alookup/alookup_test.go
vishalmohanty: Removed `STATUS_NO_ANSWER` altogether because it is not a standard dns error code. In all cases where we were returning `STATUS_NO_ANSWER`, we are now returning `NOERROR`

modules/alookup/alookup_test.go
vishalmohanty: Returning `STATUS_ERROR` in case of loops now.

vishalmohanty:
Changes made:

- Removed `STATUS_NO_ANSWER` because it is not a standard dns error code. Returning `STATUS_NOERROR` in its place.
- If IPv4 or IPv6 lookup fail with a certain error (like `SERVFAIL`), we return the same error code.
- Added UTs for verifying we return the same error code as returned by miekgdns. The idea is to make all modules consistent with the expected response from an actual dns resolver.
- Fixed few lints.

spencerdrak (commented):
I think with all the suggested changes made, things look good. I'll hold until others have the chance to also review again.

modules/alookup/alookup_test.go
```diff
@@ -23,36 +23,41 @@ import (
...
-	glf.IPv6Lookup = false
 
 	rlf := new(RoutineLookupFactory)
 	rlf.Factory = glf
 	rlf.Client = new(dns.Client)
 
 	l, err := rlf.MakeLookup()
 	if l == nil || err != nil {
-		t.Error("Failed to initialize lookup")
+		panic("Failed to initialize lookup")
```
gakiwate: @vishalmohanty I am assuming this change is to make it consistent with other lookup files?

modules/alookup/alookup_test.go
```diff
@@ -257,9 +349,140 @@ func TestDoLookup(t *testing.T) {
...
+		Protocol:    "",
+		Flags:       miekg.DNSFlags{},
+	}
+	// Verify leaf returns correctly
+	res, _, _, _ := l.DoLookup("leaf.intermediate.example.com", "")
+	verifyResult(t, res.(Result), []string{"192.0.2.3"}, nil)
+
+	// Verify empty non-terminal returns no answer
+	res, _, status, _ := l.DoLookup("intermediate.example.com", "")
+	if status != zdns.STATUS_NOERROR {
```
gakiwate, line 453: You also need to check that no answers were returned. The NOERROR does not mean no answers were returned.

modules/alookup/alookup_test.go
```diff
@@ -257,9 +349,140 @@ func TestDoLookup(t *testing.T) {
...
+		t.Errorf("Expected no results, got %v", res)
+	}
+}
+
+func TestNXDomain(t *testing.T) {
+	_, glf, _, l := InitTest()
+	glf.IPv4Lookup = true
+	res, _, status, _ := l.DoLookup("nonexistent.example.com", "")
+	if status != zdns.STATUS_NXDOMAIN {
+		t.Errorf("Expected STATUS_NOERROR status, got %v", status)
```
gakiwate: typo?

modules/alookup/alookup_test.go
vishalmohanty: That is done in the next line.

modules/alookup/alookup_test.go
vishalmohanty: Yes, Thanks for the catch!

modules/alookup/alookup_test.go
vishalmohanty: Thanks for catching this. I'm doing `t.Error` now.

spencerdrak (approved):

modules/alookup/alookup.go
```diff
@@ -43,13 +44,47 @@ func (s *Lookup) DoLookup(name, nameServer string) (interface{}, zdns.Trace, zdn
 	return s.DoTargetedLookup(name, nameServer)
 }
 
+// Verify that A record is indeed IPv4 and AAAA is IPv6
+func verifyAddress(ansType string, ip string) bool {
+	isIpv4 := net.ParseIP(ip).To4() != nil
+	isIpv6 := net.ParseIP(ip).To16() != nil && strings.Contains(ip, ":")
```
codyprime: One more comment, if you want a minor optimization.  Since:
* You are always calling `To16()`
* `To16()` will guarantee, if non-nil, that `ip` is a valid IP

You don't need the call to `To4()`.  Logic like this would suffice and save a conversion call:

```go
var isIpv4, isIpv6 bool
if net.ParseIP(ip).To16() != nil {
    isIpv6 = strings.Contains(ip, ":")
    isIpv4 = !isIpv6
}
```

But even beyond that, I don't think the call to `To16()` is even needed; `ParseIP` will return `nil` if is not a valid textual representation of an IP address [1].  So I think this can get abbreviated down to:

```go
var isIpv4, isIpv6 bool
if net.ParseIP(ip) != nil {
    isIpv6 = strings.Contains(ip, ":")
    isIpv4 = !isIpv6
}
```

[1] https://pkg.go.dev/net#ParseIP

vishalmohanty: Done. Also initialised isIpv4 and isIpv6 to false.

PR #264, opened 2022-01-29
Add Versioning to ZDNS

spencerdrak:
This PR simply adds versioning to ZDNS. It will find the latest tags and increment them by one minor (patch) version each time a push to master is made. Since v1.0.0-RC1 exists already, it will begin tagging with v1.0.1 (no RC tag will be applied).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,11 @@
 name: CI
-on: [ push, pull_request ]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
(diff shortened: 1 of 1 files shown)


zakird:
I'm not sure that we want to version every time that a commit is pushed. I suspect that it makes more sense to handle this manually so that we have control of when versions go are released. @dadrian thoughts?

dadrian:
Agreed. I'd set up Goreleaser to run on manual tags instead, like we do on ZLint.

Issue #256, opened 2021-10-19
panic when retries is set to 0


yairhoff:
When I run zdns with "-retries 0" I get the following panic:
panic: loop must return

goroutine 12 [running]:
github.com/zmap/zdns/modules/miekg.(*Lookup).retryingLookup(0xc0000a0050, {0x10, 0x1, {0xc000090000, 0x8}}, {0xc00045fd10, 0x10}, 0xf)
	/home/yair/git-projects/zdns/modules/miekg/miekg.go:348 +0x4a5
github.com/zmap/zdns/modules/miekg.(*Lookup).cachedRetryingLookup(0xc0000a0050, {0xfd10, 0x45, {0xc000090000, 0x10}}, {0xc00045fd10, 0x10}, {0xc000415ae0, 0x8}, 0x3)
	/home/yair/git-projects/zdns/modules/miekg/miekg.go:414 +0xf53
github.com/zmap/zdns/modules/miekg.(*Lookup).iterativeLookup(0xc0000a0050, {0xbc0, 0x8b, {0xc000090000, 0x0}}, {0xc00045fd10, 0x10}, 0x3, {0xc000415ae0, 0x8}, ...)
	/home/yair/git-projects/zdns/modules/miekg/miekg.go:576 +0x30f
github.com/zmap/zdns/modules/miekg.(*Lookup).iterateOnAuthorities(_, {_, _, {_, _}}, _, {{0x8b0bc0, 0x0, 0x0}, {0x8b0bc0, ...}, ...}, ...)
	/home/yair/git-projects/zdns/modules/miekg/miekg.go:549 +0x878
github.com/zmap/zdns/modules/miekg.(*Lookup).iterativeLookup(0xc0000a0050, {0xbc0, 0x8b, {0xc000090000, 0x0}}, {0xc00045fa60, 0x10}, 0x2, {0xc000415028, 0x3}, ...)
	/home/yair/git-projects/zdns/modules/miekg/miekg.go:610 +0xb48
github.com/zmap/zdns/modules/miekg.(*Lookup).iterateOnAuthorities(_, {_, _, {_, _}}, _, {{0x8b0bc0, 0x0, 0x0}, {0xc000448200, ...}, ...}, ...)
	/home/yair/git-projects/zdns/modules/miekg/miekg.go:549 +0x878
[...]

PR #265, opened 2022-01-31
Unit tests for SPF and DMARC

bayerhonza:
While working #254, I wrote few unit tests. I created a new PR as I added a new package for Go unit testing [stretchr/testify](https://github.com/stretchr/testify) and I was not sure if ZDNS maintainers agree with it.

If the addition of such a library is not welcome, I will rewrite the tests so that no other library is needed.


dadrian:
Would you be willing to switch from testify to `gotest.tools`? It has basically the same API, but is still maintained and has a couple extra features.

bayerhonza:
Ok, done.

spencerdrak:
This looks good to me. If @dadrian doesn't have any other thoughts, I will merge. The checks are passing.

modules/bindversion/bindversion.go
```diff
@@ -34,14 +34,10 @@ type Lookup struct {
...
 func (s *Lookup) DoLookup(_, nameServer string) (interface{}, zdns.Trace, zdns.Status, error) {
-	var res Result
-	res.Resolver = nameServer
-	innerRes, trace, status, err := s.DoTxtLookup("VERSION.BIND", nameServer)
-	if status != zdns.STATUS_NOERROR {
-		return res, trace, status, err
-	}
-	res.BindVersion = innerRes
-	return res, trace, zdns.STATUS_NOERROR, nil
+	innerRes, trace, status, err := s.DoLookup("VERSION.BIND", nameServer)
```
vishalmohanty: Isn't this calling the same `DoLookup`? Infinite loop?

vishalmohanty:
Also @bayerhonza can you add UTs for bindversion?

modules/bindversion/bindversion.go
bayerhonza: Yes it is obviously a bug, I fixed it

bayerhonza:
I've added UTs for bind and improved UTs for SPF and DMARC. Also, I initially created this PR to separate UTs from the implementation itself that is discussed in #254. Shall I close this one and merge everything in #254 or are you going to merge this PR #256 and close the other?

spencerdrak:
Merged 254 which included this, so I think we're all good. Thanks!

PR #267, opened 2022-02-01
Break CLI from ZDNS Library, Add `cobra`/`viper` support

spencerdrak:
Apologies for the rather large PR, I attempted to scope this as much as possible. A summary of changes and future work is below:

Changes:
- Use `spf13/cobra` as the CLI framework
- Use `spf13/viper` to allow ENV/config file settings for flags
- Use `spf13/pflag` for cleaner integration across packages
- Move CLI into the main package
- Move library into `github.com/zmap/zdns/pkg/zdns`
- Update makefile, github actions, and IT tests to reflect new changes.

Future Work (left out to keep the scope of this PR appropriate):
- Add in support for passing a name on the CLI like `dig`
- Rework the API to clean up flag passing and global config
- sort out global state declarations across the main package

I would definitely appreciate a review on this one - I'm sure there's areas that could use improvement. I've taken a pass or two myself and iterated to clean up, but, a second (or third) pair of eyes would be awesome!


zakird:
@spencerdrak can you take a look at tests

spencerdrak:
Having a bit of issue replicating the failure even with the same merging pattern. Going to have to do some deeper digging on this tomorrow and push an update once it's fixed.

.gitignore
```diff
@@ -1,4 +1,4 @@
-zdns/zdns
+zdns
```
dadrian, line 1: This will ignore the zdns folder

dadrian (approved):
Looks reasonable, but I'm not sure I follow moving ZDNS itself to `pkg/zdns/zdns`, vs leaving it in the root

.gitignore
spencerdrak: currently we don't have a `zdns` folder at the root, but, if we want to have one, we'll need to rename the binary in order to keep ignoring it. Do you have a preference on which we do?

spencerdrak:
I left some brief notes in the readme. As mentioned above, I'll be working on cleaning up the interface, at which point, I'll add more detail and usage examples to the README and/or a wiki page.

PR #268, opened 2022-02-03
Add UTs for mxlookup and cname support

vishalmohanty:
This diff does the following

- Adds cname support for mxlookup to lookup IP addresses if there are cname records.
- Adds UTs for mxlookup
- Replaces panic with nil return with error

### DIFF
--- a/modules/common/common.go
+++ b/modules/common/common.go
@@ -1,0 +1,24 @@
+package common
+
+import (
+	"net"
+	"strings"
+
+	"github.com/zmap/zdns"
+)
+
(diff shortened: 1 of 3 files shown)


zakird (changes requested):
I would suggest moving this CNAME logic into a helper utility. It's the same logic as is in the ALOOKUP module and I don't think it makes sense to duplicate here.

vishalmohanty:
Closing PR, will reopen once https://github.com/zmap/zdns/issues/269 is addressed.

PR #272, opened 2022-02-06
Updating README with right installation syntax

luca-regne:


stklcode:
If we talk about README corrections, the build instructions are also incorrect: `cd zdns/zdns` (there is no “zdns“ subdirectory anymore, should be `cd zdns` only)

spencerdrak:
Agreed - looks like we missed those spots when we made the changes. If @luca-regne wants to fix that as well, awesome, otherwise I'll open another PR to correct it shortly.

spencerdrak (approved):

PR #275, opened 2022-02-10
Modularise DoLookupProtocol

### LINKED ISSUE #269: Define doLookupProtocol in a common place so that it can be used across modules
Currently every module is having to define `doLookupProtocol` in order to get IP addresses of either CNAME records, MX records, NS records, etc. 

I was able to define `DoLookupProtocol` in `common.go` but faced issue in running unit tests like `alookup_test.go` because the mocked `DoMiekgLookup` was not being called. Instead the actual implementation in `miekg.go` was being invoked.

There should be some way of injecting dependencies or mocking functions of other modules, which will help us abstract out `doLookupProtocol`. Investigate that.

vishalmohanty:
This diff does the following

- Abstracts out the `DoLookupProtocol` from `alookup` so that it can be used by other modules.
- Add UTs for `DoLookupProtocol` by mocking the `DoMiekgLookup` using Interface.

### DIFF
--- a/pkg/miekg/miekg.go
+++ b/pkg/miekg/miekg.go
@@ -770,15 +770,19 @@ func (s *Lookup) DoTargetedLookup(l LookupClient, name, nameServer string, looku
 
 	if lookupIpv4 {
 		ipv4, ipv4Trace, ipv4status, _ = s.DoIpsLookup(l, name, nameServer, dns.TypeA, candidateSet, cnameSet, name, 0)
-		res.IPv4Addresses = make([]string, len(ipv4))
-		copy(res.IPv4Addresses, ipv4)
+		if len(ipv4) > 0 {
+			res.IPv4Addresses = make([]string, len(ipv4))
+			copy(res.IPv4Addresses, ipv4)
+		}
 	}
 	candidateSet = map[string][]Answer{}
 	cnameSet = map[string][]Answer{}
 	if lookupIpv6 {
 		ipv6, ipv6Trace, ipv6status, _ = s.DoIpsLookup(l, name, nameServer, dns.TypeAAAA, candidateSet, cnameSet, name, 0)
-		res.IPv6Addresses = make([]string, len(ipv6))
-		copy(res.IPv6Addresses, ipv6)
+		if len(ipv6) > 0 {
+			res.IPv6Addresses = make([]string, len(ipv6))
+			copy(res.IPv6Addresses, ipv6)
+		}
 	}
 
 	combinedTrace := append(ipv4Trace, ipv6Trace...)
@@ -791,7 +795,7 @@ func (s *Lookup) DoTargetedLookup(l LookupClient, name, nameServer string, looku
 		} else if lookupIpv6 && !SafeStatus(ipv6status) {
 			return nil, combinedTrace, ipv6status, nil
 		} else {
-			return nil, combinedTrace, zdns.STATUS_NOERROR, nil
+			return res, combinedTrace, zdns.STATUS_NOERROR, nil
 		}
 	}
 	return res, combinedTrace, zdns.STATUS_NOERROR, nil
--- a/pkg/miekg/miekg.go
+++ b/pkg/miekg/miekg.go
@@ -687,29 +687,13 @@ func (s *Lookup) DoMiekgLookup(q Question, nameServer string) (interface{}, zdns
 	}
 }
 
-// Verify that A record is indeed IPv4 and AAAA is IPv6
-func verifyAddress(ansType string, ip string) bool {
-	isIpv4 := false
-	isIpv6 := false
-	if net.ParseIP(ip) != nil {
-		isIpv6 = strings.Contains(ip, ":")
-		isIpv4 = !isIpv6
-	}
-	if ansType == "A" {
-		return isIpv4
-	} else if ansType == "AAAA" {
-		return isIpv6
-	}
-	return !isIpv4 && !isIpv6
-}
-
 func populateResults(records []interface{}, dnsType uint16, candidateSet map[string][]Answer, cnameSet map[string][]Answer, garbage map[string][]Answer) {
 	for _, a := range records {
 		// filter only valid answers of requested type or CNAME (#163)
 		if ans, ok := a.(Answer); ok {
 			lowerCaseName := strings.ToLower(ans.Name)
 			// Verify that the answer type matches requested type
-			if verifyAddress(ans.Type, ans.Answer) {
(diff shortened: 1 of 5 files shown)


pkg/miekg/miekg_test.go
```diff
@@ -1,12 +1,327 @@
...
+
+var status = zdns.STATUS_NOERROR
+
+type MockLookupClient struct{}
+
+func (mc MockLookupClient) ProtocolLookup(s *Lookup, q Question, nameServer string) (interface{}, zdns.Trace, zdns.Status, error) {
+	if res, ok := mockResults[q.Name]; ok {
+		return res, nil, status, nil
+	} else {
+		return nil, nil, zdns.STATUS_NXDOMAIN, nil
```
spencerdrak, line 25: Are there any other statuses that we may want to return here and see how the code handles it? STATUS_NXDOMAIN seems like it'd be the most common, but maybe in some corner cases we'd get something else.

pkg/miekg/miekg.go
```diff
@@ -650,6 +663,93 @@ func (s *Lookup) DoMiekgLookup(q Question, nameServer string) (interface{}, zdns
...
+					garbage[lowerCaseName] = append(garbage[lowerCaseName], ans)
+				}
+			} else {
+				garbage[lowerCaseName] = append(garbage[lowerCaseName], ans)
+			}
+		}
+	}
+}
+
+func (s *Lookup) DoProtocolLookup(lc LookupClient, name string, nameServer string, dnsType uint16, candidateSet map[string][]Answer, cnameSet map[string][]Answer, origName string, depth int) ([]string, []interface{}, zdns.Status, error) {
```
zakird: If I'm reading this code correctly, this function does A/AAAA lookups with CNAME following, correct? I suspect that we can come up with a better function name that makes this clearer. The current name is very generic.

pkg/miekg/miekg.go
```diff
@@ -1,3 +1,6 @@
+//go:build !mock
```
spencerdrak: What's the uses for the build constraints here? Do we need to pass any flags to build mocks while testing?

spencerdrak: rather, does this need to be excluded from any builds during testing?

pkg/miekg/miekg.go
```diff
@@ -74,6 +77,16 @@ type GlobalLookupFactory struct {
 	BlMu           sync.Mutex
 }
 
+// Lookup client interface for helping in mocking
+type LookupClient interface {
+	ProtocolLookup(s *Lookup, q Question, nameServer string) (interface{}, zdns.Trace, zdns.Status, error)
+}
+
+// Check whether the status is safe
+func SafeStatus(status zdns.Status) bool {
```
spencerdrak: If this is used anywhere else (I'm not sure if the modules will have any use for this, or the CLI) it may be better defined in the util package.

spencerdrak (commented):
I'd be curious if anyone else has feedback on how we've chosen to structure the mocking capability though. I think it's sensible, but another set of eyes would be good.

pkg/miekg/miekg_test.go
vishalmohanty: We have the `status` global variable for that. Look at `TestServFail`

pkg/miekg/miekg.go
vishalmohanty: Sorry, this was from another thing I was trying. Removed now

pkg/miekg/miekg.go
vishalmohanty: Moved to util

pkg/miekg/miekg.go
vishalmohanty: Yes. I changed it to `DoIpsLookup`

vishalmohanty:
Summary of changes

- Moved `DoTargetedLookup` out of `alookup` as that can be utilised by other modules as well.
- Moved the UTs from `alookup` to `miekg` as all the logic now resides in `miekg`.
- Added basic UTs in `alookup` to ensure the flags are read correctly.

spencerdrak (approved):
@vishalmohanty this looks good to me. We may have some additional work to expand tests further in the future, but this looks good to me!

pkg/miekg/miekg_test.go
```diff
@@ -1,13 +1,483 @@
...
+		Additional:  nil,
+		Authorities: nil,
+		Protocol:    "",
+		Flags:       DNSFlags{},
+	}
+	res, _, _, _ := a.DoTargetedLookup(mc, "example.com", gc.NameServers[0], true, false)
+	verifyResult(t, res.(IpResult), []string{"192.0.2.1", "192.0.2.2"}, nil)
+}
+
+func TestQuadAWithoutFlag(t *testing.T) {
```
gakiwate, line 99: What is this function trying to test? 

Also, generally could we add a line to each of the test cases as to the goal of the UT?

pkg/miekg/miekg_test.go
```diff
@@ -1,13 +1,483 @@
...
+		Additional:  nil,
+		Authorities: nil,
+		Protocol:    "",
+		Flags:       DNSFlags{},
+	}
+	res, _, _, _ := a.DoTargetedLookup(mc, "example.com", gc.NameServers[0], false, true)
+	verifyResult(t, res.(IpResult), nil, []string{"2001:db8::1", "2001:db8::2"})
+}
+
+func TestNoResults(t *testing.T) {
```
gakiwate, line 203: What does it mean for there to be no answers and the status to be NOERROR? Is this the empty non terminal case?

pkg/miekg/miekg_test.go
```diff
@@ -1,13 +1,483 @@
...
+		Additional:  nil,
+		Authorities: nil,
+		Protocol:    "",
+		Flags:       DNSFlags{},
+	}
+	res, _, _, _ := a.DoTargetedLookup(mc, "cname.example.com", gc.NameServers[0], false, true)
+	verifyResult(t, res.(IpResult), nil, []string{"2001:db8::3"})
+}
+
+func TestUnexpectedMxOnly(t *testing.T) {
```
gakiwate, line 280: What is being tested here?

vishalmohanty: Added comment

pkg/miekg/miekg_test.go
vishalmohanty: Added comments

pkg/miekg/miekg_test.go
vishalmohanty: Updated to return empty result in case of `NOERROR`

PR #276, opened 2022-02-10
Update mxlookup and add UTs

vishalmohanty:
This diff does the following

- Adds cname support for mxlookup to lookup IP addresses if there are cname records.
- Adds UTs for mxlookup
- Replaces panic with nil return with error

### DIFF
--- a/pkg/mxlookup/mxlookup_test.go
+++ b/pkg/mxlookup/mxlookup_test.go
@@ -61,6 +61,7 @@ type minimalServerRecords struct {
 
 func InitTest(t *testing.T) (*zdns.GlobalConf, *GlobalLookupFactory, *RoutineLookupFactory, zdns.Lookup) {
 	mxResults = make(map[string]miekg.Result)
+	mockResults = make(map[string]miekg.IpResult)
 	gc := new(zdns.GlobalConf)
 	gc.NameServers = []string{"127.0.0.1"}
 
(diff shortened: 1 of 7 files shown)


spencerdrak:
@vishalmohanty will merging in 275 beneath this PR ensure that all these tests pass, or was something else misconfigured there?

vishalmohanty:
@spencerdrak there was an issue in code. I fixed it now. Tests are passing.

pkg/mxlookup/mxlookup_test.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+type minimalServerRecords struct {
+	recType       string
+	IPv4Addresses []string
+	IPv6Addresses []string
+}
+
+func InitTest(t *testing.T) (*zdns.GlobalConf, *GlobalLookupFactory, *RoutineLookupFactory, zdns.Lookup) {
```
spencerdrak, line 65: Do we want to have any UTs for failure modes in this module?

spencerdrak (commented):
Looks good except we may want to expand testing as noted.

pkg/mxlookup/mxlookup_test.go
vishalmohanty: Thanks for this. Added.

spencerdrak (approved):
I think this looks good - Gautam may have thoughts on the test cases, though.

PR #277, opened 2022-02-10
Update nslookup and add UTs

vishalmohanty:
This diff does the following

- Adds support for following CNAME paths recursively in NS lookups to get IPs
- Verification of IP address and type match in Additionals section.
- Adds UTs for nslookup

### DIFF
--- a/pkg/nslookup/nslookup_test.go
+++ b/pkg/nslookup/nslookup_test.go
@@ -10,13 +10,12 @@ import (
 )
 
 var nsRecords = make(map[string]miekg.Result)
-
-var errCode = zdns.STATUS_NOERROR
+var miekgStatus = zdns.STATUS_NOERROR
 
 // Mock the actual Miekg lookup.
 func (s *Lookup) DoMiekgLookup(question miekg.Question, nameServer string) (interface{}, []interface{}, zdns.Status, error) {
 	if res, ok := nsRecords[question.Name]; ok {
-		return res, nil, errCode, nil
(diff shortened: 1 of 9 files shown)


pkg/nslookup/nslookup_test.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil, nil, zdns.STATUS_NXDOMAIN, nil
+	}
+}
+
+type minimalServerRecords struct {
+	IPv4Addresses []string
+	IPv6Addresses []string
+}
+
+func InitTest(t *testing.T) (*zdns.GlobalConf, *GlobalLookupFactory, *RoutineLookupFactory, zdns.Lookup) {
```
spencerdrak, line 50: Here as well, we'll likely want to test some failure modes. Additionally, we may be able to abstract some of that away to the util package if the test logic is shared across modules

spencerdrak (commented):
Looks good except we may want to expand testing as noted.

pkg/nslookup/nslookup_test.go
vishalmohanty: Added failure modes. I don't want to move `InitTest` to util since `GlobalLookupFactory` and `RoutineLookupFactory` differs across modules.

spencerdrak: Concur, I must've overlooked the differences.

spencerdrak (approved):
Here as well, I think this is good. Gautam may have some thoughts as well, though.

PR #281, opened 2022-02-11
Refactor interface between end-user (CLI) and library

spencerdrak:
This PR further decouples the ZDNS library from the CLI. A few key changes are outlined below:

- configuration is all moved into a `ZdnsRun` structure for ease of use
- defaults are set, and warnings emitted, when fields are not appropriately set
- Setting flags for modules has changed slightly, to decouple the CLI from the lib. More changes to this will likely occur.
- Appropriate updates and example in the README 
- Brings back dig-like syntax

Other changes (per discussions with @zakird) will include further breaking up what is currently held in the library to clarify how modules should look. These changes were left out of the PR so that the changes weren't overwhelming.

All tests pass and checks are successful.


spencerdrak:
Update to the PR - I also included dig-like syntax here, since many of the changes were related/overlapping.

PR #282, opened 2022-02-16
Refactor interface between end-user (CLI) and library

spencerdrak:
This PR further decouples the ZDNS library from the CLI. A few key changes are outlined below:

- configuration is all moved into a ZdnsRun structure for ease of use
- generalizes module input, no special cases for lookup modules
- defaults are set, and warnings emitted, when fields are not appropriately set
- Setting flags for modules has changed slightly, to decouple the CLI from the lib. More changes to this will likely occur.
- Appropriate updates and example in the README

Other changes (per discussions with @zakird) will include further breaking up what is currently held in the library to clarify how modules should look. There's a few TODOs in here that relate to this as well. These changes were left out of the PR so that the changes weren't overwhelming.

All tests pass and checks are successful.


zakird:
I'm not sure if it makes sure to turn this type of logic into the library
versus just the recursion + caching, etc. What's the use case of someone using
our Go Routines, etc.?  By the time you figure out how to use all of our
GlobalConf and such, it's probably a lot more work than simply using the
primitives that we provide directly. It similarly doesn't seem correct to me
to have a library that accesses something like `stdin` and not have an interface.
it might make sense to have a quick meeting to discuss the interface.

@paul-pearce , @dadrian , others, what are your thoughts?

README.md
```diff
@@ -37,9 +37,56 @@ Usage
...
+
+	// Configure this run to use the alookup module.
+	run.GlobalConf.Module = "ALOOKUP"
+
+	// Set options for all goroutines
+	run.Class = "INET"
+
+	// Run the lookup
+	zdns.Run(run)
+}
```
zakird, line 70: I'm not sure that this makes sense. We shouldn't have a library that blindly takes things in from `STDIN` I don't think. I believe that the library—if anything should take in a channel or such—that the CLI processes.

spencerdrak: That's a fair point I hadn't considered, I can adjust this. We can discuss the rest of those points at our meeting so I have a clear picture. Thanks!

PR #289, opened 2022-02-24
Add AXFR unit tests

### LINKED ISSUE #279: Test AXFR module
The AXFR module requires `allow-transfer` directive to dig against a nameserver. Figure out how to test it via unit and integration tests.

vishalmohanty:
This diff does the following

- Add UT coverage for AXFR.
- Add the `dns.Transfer` object to the `Lookup` object in `axfr` to be able to mock it for the UT.
- Other minor edits to axfr.go

### DIFF
--- a/pkg/axfr/axfr.go
+++ b/pkg/axfr/axfr.go
@@ -34,6 +34,7 @@ import (
 type Lookup struct {
 	Factory *RoutineLookupFactory
 	nslookup.Lookup
+	dns.Transfer
 }
 
 type AXFRServerResult struct {
@@ -51,6 +52,10 @@ func dotName(name string) string {
 	return strings.Join([]string{name, "."}, "")
 }
 
+type TransferClient struct {
+	dns.Transfer
+}
+
 func (s *Lookup) DoAXFR(name, server string) AXFRServerResult {
 	var retv AXFRServerResult
 	retv.Server = server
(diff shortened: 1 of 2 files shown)


pkg/axfr/axfr.go
```diff
@@ -59,32 +64,31 @@ func (s *Lookup) DoAXFR(name, server string) AXFRServerResult {
...
 		if blacklisted, err := s.Factory.Factory.Blacklist.IsBlacklisted(server); err != nil {
 			s.Factory.Factory.BlMu.Unlock()
-			retv.Status = "ERROR"
+			retv.Status = string(zdns.STATUS_ERROR)
 			retv.Error = "blacklist-error"
 			return retv
 		} else if blacklisted {
 			s.Factory.Factory.BlMu.Unlock()
-			retv.Status = "ERROR"
+			retv.Status = string(zdns.STATUS_ERROR)
```
spencerdrak: Curious if there's a specific reason to keep the status and other fields in the `AXFRServerResult` as strings, not some sort of typed value.

spencerdrak (commented):
I think this looks good - but likely Gautam will have some thoughts, he'll be more familiar with what we should be testing.

pkg/axfr/axfr.go
vishalmohanty: I don't see an issue with having the Status as type `zdns.Status`. @zakird any reason it was initially written to be `string`?

zakird: I don't think so, makes a lot of sense to me to use `zdns.Status`.

vishalmohanty: Updated to use `zdns.Status`

gakiwate:
@spencerdrak Looks good to me.

spencerdrak (approved):

PR #290, opened 2022-02-24
Parse answer tests for mx, ns, spf

### LINKED ISSUE #280: Complete TestParseAnswer in miekg_test.go
TestParseAnswer is a UT for testing that parsing of records of various kinds works successfully. There's a TODO in the UT to add test other records as well. Check if we need to test for other records.

vishalmohanty:
This diff adds parsing tests for mx, ns, spf which are the important modules from the point of zdns use case.

### DIFF
--- a/pkg/miekg/miekg_test.go
+++ b/pkg/miekg/miekg_test.go
@@ -631,7 +631,48 @@ func TestParseAnswer(t *testing.T) {
 		t.Errorf("Unxpected replacement. Expected %v, got %v", ".", answer.Replacement)
 	}
 
-	// TODO: test remaining RR types
+	// MX record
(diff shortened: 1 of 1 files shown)


spencerdrak (commented):
Looks good - however, are there any failure modes we want to test here? Might be good to establish under what conditions we expect a failure.

vishalmohanty:
Added negative test scenarios when cname record is returned when A and AAAA is expected. @spencerdrak or Gautam any other scenarios you think we should cover?

spencerdrak:
per separate convo with Gautam, we're good to merge this one as well.

spencerdrak (approved):

PR #291, opened 2022-02-24
Add integration tests for AXFR

### LINKED ISSUE #279: Test AXFR module

vishalmohanty:
For AXFR, we are using `zonetransfer.me` as that provides AXFR response. We check for few specific records (one from each category like A, AAAA, SOA, SPF ...) in the output from zdns.

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -6,6 +6,7 @@ import json
 import unittest
 import tempfile
 
+
 def recursiveSort(obj):
 
     def listSort(l):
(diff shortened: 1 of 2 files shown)


resources/axfr.json
```diff
@@ -0,0 +1,113 @@
+{
```
spencerdrak, line 1: I think this is a useful pattern to follow, especially as the number of tests grows.

spencerdrak: this pattern meaning having the results in a resources file.

integration_tests.py
```diff
@@ -708,6 +744,20 @@ def test_dmarc_iterative(self):
         self.assertSuccess(res, cmd)
         self.assertEqual(res["data"], self.DMARC_ANSWER["data"])
 
+    def test_axfr(self):
+        # In this test, we just check for few specific records
```
spencerdrak, line 748: Is there any specific method for choosing which of the records we keep and which we discard?

spencerdrak:
Per Gautam and @vishalmohanty, we're good to merge.

spencerdrak (approved):

PR #296, opened 2022-03-28
Add flag for all NS lookup

vishalmohanty:
This diff does the following

- Adds a flag "all-nameservers" that can be used to perform the query (like A, AAAA, MX, ..) via all the nameservers of the domain.
- Adds checks so that "all-nameservers" mode can be used when no explicit nameservers are specified.
- Moves the logic of `DoNsLookup` to `miekg` because that is being used by multiple places - `nslookup` module and `miekg` now because of the "all-nameservers" mode.
- Moves the UTs of `nslookup_test.go` to `miekg_test.go`. Delete the `nslookup_test.go` because it is no longer needed.


spencerdrak (commented):
@vishalmohanty we'll probably want to also update the README and any other docs that pertain to this. I think the structure is good, but we'll need Gautam to review as well.

gakiwate:
> @vishalmohanty we'll probably want to also update the README and any other docs that pertain to this. I think the structure is good, but we'll need Gautam to review as well.

Looks good to me.

spencerdrak (approved):

PR #306, opened 2022-05-05
Check ulimit on supported plattforms

smeinecke:
Add check if maximum connections of go processes * threads settings exceed ulimit of current session. 
Increase limit if possible. Only added on darwin + linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-zdns
 .*
 *.iml
 *.code-workspace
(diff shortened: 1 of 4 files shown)


spencerdrak (commented):
@zakird I saw some traffic a few weeks ago about something similar, though I think this is a more permanent solution. Just wanted to make sure this doesn't conflict with that (I don't believe it does). Otherwise this looks good imo.

zakird:
I just updated docs. If this looks good let's merge.

spencerdrak (approved):

Issue #318, opened 2022-11-08
Add flag to include DNSSEC additional records in output


secastro:
More than an issue, this is a feature request, if it doesn't exist.

Is it possible to add a flag that will save, as an additional record, DNSSEC-related records when the destination nameserver is doing validation?

We run a large-scale collection for a ccTLD, and saving the responses and the RRSIG records could allow us to validate those signatures offline.

phillip-stephens:
Looks like this was resolved by #323 

When I compare `dig +dnssec` to `./zdns --dnssec` I'm getting the same response when querying for `cloudflare.com`

### dig
```
;; QUESTION SECTION:
;cloudflare.com.                        IN      A

;; ANSWER SECTION:
cloudflare.com.         300     IN      A       104.16.132.229
cloudflare.com.         300     IN      A       104.16.133.229
cloudflare.com.         300     IN      RRSIG   A 13 2 300 20240718162838 20240716142838 34505 cloudflare.com. 0bjO17rcDVjSNVLZGRCzc692BQeAHkP2pirKSPVMbnmec1ryo2AFQkyb znaXlpYxestJTpxzeqmdjLQUMWU1gA==
```

### zdns
```
$ echo "cloudflare.com" | ./zdns A --dnssec | jq
{
  "data": {
    "additionals": [
      {
        "flags": "do",
        "type": "EDNS0",
        "udpsize": 1232,
        "version": 0
      }
    ],
    "answers": [
      {
        "answer": "104.16.133.229",
        "class": "IN",
        "name": "cloudflare.com",
        "ttl": 125,
        "type": "A"
      },
      {
        "answer": "104.16.132.229",
        "class": "IN",
        "name": "cloudflare.com",
        "ttl": 125,
        "type": "A"
      },
      {
        "algorithm": 13,
        "class": "IN",
        "expiration": "20240718162846",
        "inception": "20240716142846",
        "keytag": 34505,
        "labels": 2,
        "name": "cloudflare.com",
        "original_ttl": 300,
        "signature": "GhQaXcxq+BmqR8ap8H4E604+80a0KaOSfLlYABo9rCC1OdFQhZXWQznntsccYgTn83VF/ct1HEpEcxcB6NpChg==",
        "signer_name": "cloudflare.com.",
        "ttl": 125,
        "type": "RRSIG",
        "type_covered": 1
      }
    ],
    "protocol": "udp",
    "resolver": "1.1.1.1:53"
  },
  "name": "cloudflare.com",
[...]

PR #323, opened 2023-01-16
enable EDNS0; Add --dnssec option

stirnim:
This pull request enables EDNS0 by default with an UDP Size of 1232. This should improve performance slightly as fewer truncated answers are to be expected and so less retries over TCP are required.

It also add the command-line option `--dnssec` for those who want/need DNSSEC records in the answer.

This should resolve the issues:
- https://github.com/zmap/zdns/issues/318
- https://github.com/zmap/zdns/issues/220

It should also make the following pull request obsolete:
- https://github.com/zmap/zdns/pull/221

Note:
- I did not implement OPT pseude-RR parsing and left a TODO. Can be implemented later if someone needs this.
- I noticed that NXDOMAIN responses do not return the SOA record in the authority section. Likewise, no additional DNSSEC records are returned if the `--dnssec` option is used. I guess there is some code which already filters this out. As this pull request does not change the existing behaviour, I did not attempt to change this.

### DIFF
--- a/pkg/miekg/answers.go
+++ b/pkg/miekg/answers.go
@@ -71,6 +71,14 @@ type DSAnswer struct {
 	DigestType uint8  `json:"digest_type" groups:"short,normal,long,trace"`
 	Digest     string `json:"digest" groups:"short,normal,long,trace"`
 }
+
+type EDNSAnswer struct {
+	Type    string `json:"type" groups:"short,normal,long,trace"`
+	Version uint8  `json:"version" groups:"short,normal,long,trace"`
+	Flags   string `json:"flags" groups:"short,normal,long,trace"`
+	UDPSize uint16 `json:"udp_size" groups:"short,normal,long,trace"`
+}
+
 type GPOSAnswer struct {
 	Answer
 	Longitude string `json:"preference" groups:"short,normal,long,trace"`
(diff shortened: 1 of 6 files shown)


pkg/miekg/answers.go
```diff
@@ -796,6 +796,10 @@ func ParseAnswer(ans dns.RR) interface{} {
 		return makeSVCBAnswer(&cAns.SVCB)
 	case *dns.SVCB:
 		return makeSVCBAnswer(cAns)
+	case *dns.OPT:
```
zakird, line 824: Just curious about adding this here. Is this better off than the default case where you get _some_ data back? This seems to reduce the amount of data made accessible (unless it's triggering some error condition).

stirnim: I thought the default presentation of the OPT record was not very useful but reading [draft-peltan-edns-presentation-format-00.html](https://www.ietf.org/id/draft-peltan-edns-presentation-format-00.html), the default is not far off the example given in section 3.

Instead of reverting that change, I thought I may as well add some code to parse a basic EDNS0 answer. e.g.

```json
{
"flags": "do",
"type": "EDNS0",
"udp_size": 1232,
"version": 0
}
```

zakird (approved):

pkg/miekg/answers.go
zakird: Love it. Thanks for doing that.

Issue #325, opened 2023-03-29
Allow setting the CD bit to ignore DNSSEC validation


bayerhonza:
Some public resolvers such as Google Public DNS (8.8.8.8) or Cloudflare DNS (1.1.1.1) perform DNSSEC validation. As a result, domains that are not signed return SERVFAIL.

example: 
```bash
echo "dnssec-failed.org" | zdns --name-servers 1.1.1.1 a
``` 
results to
```json
{"data":{"protocol":"udp","resolver":"1.1.1.1:53"},"name":"dnssec-failed.org","status":"SERVFAIL","timestamp":"2023-02-29T12:00:00+02:00"}
```

However, when performing DNS scans where we do not want to take into consideration the DNSSEC validation result, it is useful to set the CD bit ([RFC4035](https://www.rfc-editor.org/rfc/rfc4035#section-3.2.2)) to disable signature validation and receive raw data from the resolver.

The expected result for
```bash
echo "dnssec-failed.org" | zdns a --name-servers 1.1.1.1 --checking-disabled
``` 

would then be:
```json
{"data":{"additionals":[{"flags":"","type":"EDNS0","udpsize":1232,"version":0}],"answers":[{"answer":"96.99.227.255","class":"IN","name":"dnssec-failed.org","ttl":300,"type":"A"}],"protocol":"udp","resolver":"1.1.1.1:53"},"name":"dnssec-failed.org","status":"NOERROR","timestamp":"2023-02-29T12:00:00+02:00"}
```

It is not an issue, but rather a feature request and my pull request will follow shortly.

Thx

PR #326, opened 2023-03-29
Introducing checkingDisabled option

bayerhonza:
Implementation of #325


zakird:
This looks great. Any chance you have a configuration that we could use for a test case?

bayerhonza:
Yes, the example that I put in the issue can be used as the test case. I've added two test cases. Let me know if you want to add some UT or if the integration test will do the job.

zakird (approved):

zakird:
That's perfect — thank you

PR #327, opened 2023-04-02
Ulimit check is incorrect

bayerhonza:
Hello,

in previous versions, I could run ZDNS with tens of thousands of threads without problem on a server with 64 logical cores. However, now it is impossible because of the added ulimit check that is too strict.

The default ulimit -n value for a newly created bash is 1024. When running zdns with 1000 threads, it now shows a warning:
```
WARN[0000] Current nofile limit (1024) lower than maximum connection count (64000), try to update. 
```
with the calculated maximum connection count at 64,000 (supposedly 64 cores x 1000 threads)

 However, the number of open sockets is 1001:
```bash
printf "google.com\nexample.com"> test.txt; strace -f ./zdns A --threads 1000 --recycle-sockets=true  --input-file test.txt 2>&1 | grep 'socket(AF_INET, SOCK_DGRAM' | wc -l; rm test.txt >/dev/null
```
one socket open to get the local address: https://github.com/zmap/zdns/blob/c8e1d8e2fd7a818dd9e95d5aecf9584c34aaf703/pkg/zdns/zdns.go#L200-L206

and 1000 sockets correspond to one socket open per goroutine. The total number of open files is less than 1024 and zdns should run without problem. 

The problem IMO is coming from the maximum ulimit value:
https://github.com/zmap/zdns/blob/c8e1d8e2fd7a818dd9e95d5aecf9584c34aaf703/pkg/zdns/zdns.go#L217-L222

If I understand well GOMAXPROCS, it determines how many routines can run in parallel but it does not multiply the number of open sockets. 

I propose to set the ulimit value to the number of threads as each thread is supposed to open one socket (reusable or not).

### DIFF
--- a/pkg/zdns/ulimit_check.go
+++ b/pkg/zdns/ulimit_check.go
@@ -22,18 +22,18 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func ulimit_check(max_open_files uint64) {
+func ulimitCheck(maxOpenFiles uint64) {
 	var rLimit syscall.Rlimit
 	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
 	if err != nil {
 		log.Fatal("Failed to fetch ulimit ", err)
 	}
 
-	if max_open_files > rLimit.Cur {
-		log.Warn("Current nofile limit (", rLimit.Cur, ") lower than maximum connection count (", max_open_files, "), trying to update.")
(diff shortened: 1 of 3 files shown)


zakird (approved):

zakird:
Thanks for the fix. Looking back, I'm not sure how we arrived at that original logic, but it's certainly not right.

PR #330, opened 2023-05-05
Allow for escaped commas in CLI input

### LINKED ISSUE #329: Passing domains that have a comma as part of the string
Is it possible to submit a request for a domain that has a comma as part of the domain name ie abc.edg,hij.com . When I pass this it treats the part of the string after the comma as the resolver to be used.

thanks

dadrian:
Fixes #329

You're in luck, I was on a plane with Wifi and time to spare.


zakird (approved):

pkg/zdns/lookup.go
```diff
@@ -68,7 +69,11 @@ func parseMetadataInputLine(line string) (string, string) {
 }
 
 func parseNormalInputLine(line string) (string, string) {
-	s := strings.SplitN(line, ",", 2)
+	r := csv.NewReader(strings.NewReader(line))
+	s, err := r.Read()
+	if err != nil || len(s) == 0 {
+		return line, ""
+	}
 	if len(s) == 1 {
```
zakird, line 77: This feels like it is duplicative of L74?

pkg/zdns/lookup.go
```diff
@@ -68,7 +69,11 @@ func parseMetadataInputLine(line string) (string, string) {
 }
 
 func parseNormalInputLine(line string) (string, string) {
-	s := strings.SplitN(line, ",", 2)
+	r := csv.NewReader(strings.NewReader(line))
```
zakird, line 72: Is this cheap enough that we're not worried about creating a new reader for every line? Docs look like expected to be reused across a whole file.

PR #388, opened 2024-06-12
Add bug template for issues

phillip-stephens:


zakird (changes requested):
I think that we should extend this to support submitting both bugs and feature requests since there's active development against this project.

phillip-stephens:
@zakird Good call, I've added a `feature-request` issue template!

zakird (approved):

Issue #304, opened 2022-04-28
Lookup via IPv6 nameservers


vishalmohanty:
What is the right way to lookup via IPv6 nameservers? Once this is figured out, add the functionality to `all-nameservers` mode so that we can lookup via both IPv4 and IPv6 nameservers.

mutax:
When I was looking through the code it seems to me that no attempt is made to resolve the IPv6 adresses of NS records, so IPv6 is currently not supported when recursing locally.

paul-pearce:
@vishalmohanty @mutax I'm working on this on the fix-ipv6 branch. I have initial support committed. It appears functional but I haven't had time to fully test it yet. Please let me know what you think. zdns should now:

* Properly detect and classify v6 nameservers
* Properly detect and classify v6 local addresses (and ignore link-local)
* Based on the nameserver it's trying to connect too, utilize either established v4 and v6 state (if configured)

There's a bug where if you try to force zdns into a bad state (e.g., force bind to v4 but force v6 nameservers) it will error, which I will fix shortly.

zakird:
@paul-pearce, can you put this code in a branch? I'd like to have @phillip-stephens take a look at getting this fixed this fall.

phillip-stephens:
@zakird Just want to confirm, should I be ignoring all link-local nameservers for IPv6? Was unsure why that'd be different from IPv4.

zakird:
I think that makes sense

phillip-stephens:
Closed with merging in IPv6 support in #410

PR #389, opened 2024-06-13
Split apart CLIConfig validation, fix error where localAddress was not copied to Resolver, and add some unit tests

### LINKED ISSUE #337: zdns cannot connect to loopback resolvers without also specifying a loopback local-addr
zdns will attempt to connect to loopback using a non-loopback IP if you don't either manually specify local-addr or interface=lo. This is due to selecting a random address from gc.LocalAddrs, which by default is only populated with non-loopback IPs. This results in attempting to dial a connection from non-loopback->loopback, which fails. 

This is especially problematic in environments where a loopback resolver exists and no arguments are specified, as it auto detects a loopback resolver but not a loopback source IP.

See the following output. 

```
$:~/zdns$ echo example.com | ./zdns A --verbosity=5 --timeout=1 --retries=0
INFO[0000] No name servers specified. will use: 127.0.0.53:53 
DEBU[0000] THREADID 000005,DEPTH 01:    [****WIRE LOOKUP***  A   example.com   127.0.0.53:53] 
INFO[0001] Err is: read udp 67.x.x.x:36916: i/o timeout 
{"data":{"protocol":"udp","resolver":"127.0.0.53:53"},"name":"example.com","status":"TIMEOUT","timestamp":"2023-06-29T11:31:38-07:00"}

$:~/zdns$ echo example.com | ./zdns A --verbosity=5 --timeout=1 --retries=0 --local-interface=lo
INFO[0000] No name servers specified. will use: 127.0.0.53:53 
INFO[0000] using local interface: 0xc24d10              
DEBU[0000] THREADID 000003,DEPTH 01:    [****WIRE LOOKUP***  A   example.com   127.0.0.53:53] 
INFO[0000] Err is: <nil>                                
{"data":{"additionals":[{"flags":"","type":"EDNS0","udpsize":65494,"version":0}],"answers":[{"answer":"93.184.216.34","class":"IN","name":"example.com","ttl":6634,"type":"A"}],"protocol":"udp","resolver":"127.0.0.53:53"},"name":"example.com","status":"NOERROR","timestamp":"2023-06-29T11:31:45-07:00"}

$:~/zdns$ echo example.com | ./zdns A --verbosity=5 --timeout=1 --retries=0 --local-addr=127.0.0.1
INFO[0000] No name servers specified. will use: 127.0.0.53:53 
INFO[0000] using local address: 0xc24d00                
DEBU[0000] THREADID 000005,DEPTH 01:    [****WIRE LOOKUP***  A   example.com   127.0.0.53:53] 
INFO[0000] Err is: <nil>                                
{"data":{"additionals":[{"flags":"","type":"EDNS0","udpsize":65494,"version":0}],"answers":[{"answer":"93.184.216.34","class":"IN","name":"example.com","ttl":6624,"type":"A"}],"protocol":"udp","resolver":"127.0.0.53:53"},"name":"example.com","status":"NOERROR","timestamp":"2023-06-29T11:31:54-07:00"}
```

Note I added the INFO logging above to identify the actual true error under the timeout... That's probably another bug.

phillip-stephens:
Closes #337 

## Changes
- Fixed bug where if `--interface-name` was provided, the `local-addr` string was used to look-up the interface
- Fixed bug where if the `--local-addr` string was provided, it wouldn't be copied correctly and used by ZDNS.
- Added code so that if a loopback name server was to be used, a loopback local address would also be used
- Throw error if both a loopback and non-loopback name server is provided. Prevents us having to handle the complexities of handling both
- Added validation for interface + nameservers need to be both loopback or neither loopback

## Testing
The original 3 queries mentioned in the Issue (#337) pass:
```
~/zdns on  phillip/337-loopback! ⌚ 3:07:29
$ echo example.com | ./zdns A --verbosity=5 --timeout=1 --retries=0
INFO[0000] No name servers specified. will use: 127.0.0.53:53
INFO[0000] using local address(es): [127.0.0.1]
DEBU[0000] DEPTH 01:    [****WIRE LOOKUP***  A   example.com   127.0.0.53:53]
{"data":{"additionals":[{"flags":"","type":"EDNS0","udpsize":65494,"version":0}],"answers":[{"answer":"93.184.215.14","class":"IN","name":"example.com","ttl":3540,"type":"A"}],"protocol":"udp","resolver":"127.0.0.53:53"},"name":"example.com","status":"NOERROR","timestamp":"2024-06-13T03:07:40Z"}

~/zdns on  phillip/337-loopback! ⌚ 3:07:40
$ echo example.com | ./zdns A --verbosity=5 --timeout=1 --retries=0 --local-interface=lo
INFO[0000] No name servers specified. will use: 127.0.0.53:53
INFO[0000] using local interface: lo
INFO[0000] using local address(es): [127.0.0.1 ::1]
DEBU[0000] DEPTH 01:    [****WIRE LOOKUP***  A   example.com   127.0.0.53:53]
{"data":{"additionals":[{"flags":"","type":"EDNS0","udpsize":65494,"version":0}],"answers":[{"answer":"93.184.215.14","class":"IN","name":"example.com","ttl":3527,"type":"A"}],"protocol":"udp","resolver":"127.0.0.53:53"},"name":"example.com","status":"NOERROR","timestamp":"2024-06-13T03:07:53Z"}

~/zdns on  phillip/337-loopback! ⌚ 3:07:53
$ echo example.com | ./zdns A --verbosity=5 --timeout=1 --retries=0 --local-addr=127.0.0.1
INFO[0000] No name servers specified. will use: 127.0.0.53:53
INFO[0000] using local address: 127.0.0.1
INFO[0000] using local address(es): [127.0.0.1]
DEBU[0000] DEPTH 01:    [****WIRE LOOKUP***  A   example.com   127.0.0.53:53]
{"data":{"additionals":[{"flags":"","type":"EDNS0","udpsize":65494,"version":0}],"answers":[{"answer":"93.184.215.14","class":"IN","name":"example.com","ttl":3518,"type":"A"}],"protocol":"udp","resolver":"127.0.0.53:53"},"name":"example.com","status":"NOERROR","timestamp":"2024-06-13T03:08:03Z"}
```


phillip-stephens:
Most of the changes here are from pulling the `ResolverConfig` validation out into separate functions. I think this both makes it easier to reason through each function and makes testing easier.

src/cli/config_validation.go
```diff
@@ -0,0 +1,224 @@
...
+
+const (
+	loopbackAddrSubnet = "127.0.0.0/8"
+	loopbackAddrString = "127.0.0.1"
+)
+
+func validateNetworkingConfig(gc *CLIConf) error {
+	// mutually exclusive CLI options
+	if gc.LocalIfaceString != "" && gc.LocalAddrString != "" {
+		return errors.New("both --local-addr and --local-interface specified.")
```
phillip-stephens:
```suggestion
		return errors.New("--local-addr and --local-interface cannot both be specified")
```

src/cli/config_validation.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	if gc.UsingLoopbackNameServer && !gc.LocalAddrSpecified {
+		// set local addr as loopback if we're using the loopback name server
+		gc.LocalAddrs = []net.IP{net.ParseIP(loopbackAddrString)}
+		gc.LocalAddrSpecified = true
+	}
+
+	if !gc.LocalAddrSpecified {
+		// Find local address for use in unbound UDP sockets
```
phillip-stephens:
```suggestion
		// Find non-loopback local address for use in unbound UDP sockets
```

src/cli/config_validation.go
```diff
@@ -0,0 +1,224 @@
...
+		if conn, err := net.Dial("udp", "8.8.8.8:53"); err != nil {
+			return fmt.Errorf("unable to find default IP address: %v", err)
+		} else {
+			gc.LocalAddrs = append(gc.LocalAddrs, conn.LocalAddr().(*net.UDPAddr).IP)
+			err := conn.Close()
+			if err != nil {
+				log.Warn("unable to close test connection to Google Public DNS: ", err)
+			}
+		}
```
phillip-stephens: Can drop else since we `return` in if

src/cli/config_validation.go
```diff
@@ -0,0 +1,224 @@
...
+		gc.ClientSubnet.SourceNetmask = uint8(netmask)
+		gc.ClientSubnet.Address = ip
+	}
+	return nil
+}
+
+func validateNameServers(gc *CLIConf) error {
+	if gc.LookupAllNameServers && gc.NameServersString != "" {
+		log.Fatal("Name servers cannot be specified in --all-nameservers mode.")
```
phillip-stephens:
```suggestion
		log.Fatal("name servers cannot be specified in --all-nameservers mode.")
```

src/cli/config_validation.go
```diff
@@ -0,0 +1,224 @@
...
+		} else {
+			ns = strings.Split(gc.NameServersString, ",")
+		}
+		for i, s := range ns {
+			nsWithPort, err := util.AddDefaultPortToDNSServerName(s)
+			if err != nil {
+				log.Fatalf("unable to parse name server: %s", s)
+			}
+			ns[i] = nsWithPort
+		}
```
phillip-stephens, line 196: Probably should check that we have at least name server after this

src/cli/config_validation.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/zmap/dns"
+
+	"github.com/zmap/zdns/src/internal/util"
+	"github.com/zmap/zdns/src/zdns"
+)
+
+const (
+	loopbackAddrSubnet = "127.0.0.0/8"
```
zakird, line 34: Do we need to handle IPv6 here too?

src/cli/config_validation.go
```diff
@@ -0,0 +1,227 @@
...
+	// Note: we rely on the value of gc.UsingLoopbackNameServer set here, so this must be called first before other validation
+	if err := validateNameServers(gc); err != nil {
+		return errors.Wrap(err, "name servers did not pass validation")
+	}
+
+	if err := validateClientSubnetString(gc); err != nil {
+		return errors.Wrap(err, "client subnet did not pass validation")
+	}
+
+	if GC.LocalAddrString != "" {
```
zakird, line 53: This was here before, but would this better be named SourceAddr? It really threw me off what LocalAddr meant when reviewing this.

src/cli/config_validation.go
```diff
@@ -0,0 +1,227 @@
...
+		log.Info("using local address: ", GC.LocalAddrString)
+		gc.LocalAddrSpecified = true
+	}
+
+	if gc.LocalIfaceString != "" {
+		li, err := net.InterfaceByName(gc.LocalIfaceString)
+		if err != nil {
+			return fmt.Errorf("invalid local interface specified: %v", err)
+		}
+		isIfaceLoopback := li.Flags&net.FlagLoopback != 0
```
zakird: 🤔

src/cli/config_validation.go
```diff
@@ -0,0 +1,227 @@
...
+		if isIfaceLoopback != gc.UsingLoopbackNameServer {
+			return fmt.Errorf("cannot mix loopback/non-loopback nameservers (%v) and interface (%s)", gc.NameServers, gc.LocalIfaceString)
+		}
+		addrs, err := li.Addrs()
+		if err != nil {
+			return fmt.Errorf("unable to detect addresses of local interface: %v", err)
+		}
+		for _, la := range addrs {
+			// strip off the network mask
+			cleanedIP := strings.Split(la.String(), "/")[0]
```
zakird: Feels like there should be a built-in for this, but not a problem either.

src/cli/config_validation.go
```diff
@@ -0,0 +1,227 @@
...
+			ip := net.ParseIP(cleanedIP)
+			if ip == nil {
+				log.Warnf("unable to parse IP address %s from interface: %s", cleanedIP, gc.LocalIfaceString)
+				continue
+			}
+			if ip.To4() == nil {
```
zakird, line 89: Why are we skipping IPv6 addresses?

src/cli/config_validation.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	if gc.UsingLoopbackNameServer && !gc.LocalAddrSpecified {
+		// set local addr as loopback if we're using the loopback name server
+		gc.LocalAddrs = []net.IP{net.ParseIP(loopbackAddrString)}
+		gc.LocalAddrSpecified = true
+	}
+
+	if !gc.LocalAddrSpecified {
+		// Find non-loopback local address for use in unbound UDP sockets
```
zakird: This no longer feels specific for unbound sockets, no?

src/cli/config_validation.go
phillip-stephens: I'll handle this when we handle adding IPv6 [support](https://github.com/zmap/zdns/issues/304). I'll add a `TODO` here tho so I don't forget to update this.

src/cli/config_validation.go
phillip-stephens: So I had the same thought, except it's called `localAddr` in the `net.Dialer` [pkg](https://pkg.go.dev/net#Dialer). I feel like for consistency we should keep it.

src/cli/config_validation.go
phillip-stephens: Lol this is a bit confusing. Let me add some comments, but we're just checking if the interface's loopback flag is set.

src/cli/config_validation.go
phillip-stephens: I'll leave a todo to come back to this when we fully handle IPv6 support.

zakird (approved):

PR #419, opened 2024-08-15
Remove ALOOKUP module and associated code

phillip-stephens:


zakird (approved):

XoMEX:
Why was this removed? I liked the usability of the alookup module. For my use-cases I mostly care about the actual IPv4/6 addresses a domain resolves to, and not the answers DNS gives me. For me, it had the following benefits:

- it resolves A and AAAA in one command
- it has an easy output structure (I can easily use `jq` to filter for `data.ipv4_addresses` and `data.ipv6_addresses`)
  - that only contains the data I care about (the actual addresses)
- it resolves CNAMEs (I guess A and AAAA do that too now)

zakird:
I think this is a fair criticism since this is a very common use case. We
shouldn't need all the same logic in ALOOKUP now that we follow CNAMEs
within the main look up path, but it is super useful to directly get the
IPs one needs.

phillip-stephens:
@zakird and @XoMEX these are fair points, I'll re-add `alookup` with the old output format. Shouldn't need much code since we have the CNAME following built in now

PR #430, opened 2024-08-23
Fix `--class` not being respected by CLI

### LINKED ISSUE #426: --class CLI option isn't respected
**Describe the bug**
In `main`, the `CLIConf.Class` (populated by `--class` is only populated into the JSON output response given to the user, it's never used to set the DNS Class field of the DNS query.

This Issue is to re-add that support

phillip-stephens:
Closes #426

NOTE - presently, all modules hard-code `INET` as the class, except `BINDVERSION` which sets it as `CHAOS`. This is not changed in this PR.

## Description
- populate lookup module's `Class` field so user's --class choice is respected

## Testing

It's not really possible (that I can figure) how to test this in integration tests. Running `tcpdump` and evaluating the output seems iffy on random CI systems and no DNS server seemed to respond when I set the class as anything other than `INET`, so I ended up just showing both in Wireshark.

With Wireshark, I ran `./zdns A zdns-testing.com --name-servers=1.1.1.1 --class=CHAOS` with the Display Filter `sudo tcpdump -i en0 udp dst port 53 and dst host 1.1.1.1 `

### `main`
```
16:57:09.209413 IP phillips-laptop.lan.56789 > one.one.one.one.domain: 53589+ [1au] A? zdns-testing.com. (45)
```

### `Phillip/426...`
```
16:57:19.112017 IP phillips-laptop.lan.60563 > one.one.one.one.domain: 14018+ [1au] A CHAOS? zdns-testing.com. (45)
```

### DIFF
--- a/src/cli/modules.go
+++ b/src/cli/modules.go
@@ -142,6 +142,7 @@ func (lm *BasicLookupModule) CLIInit(gc *CLIConf, rc *zdns.ResolverConfig) error
 	}
 	lm.LookupAllNameServers = rc.LookupAllNameServers
 	lm.IsIterative = gc.IterativeResolution
+	lm.DNSClass = gc.Class
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


phillip-stephens:
@zakird WDYT, should we explicitly warn users with non-basic modules that `--class` is not supported or should I respect `--class` no matter the module?

zakird (approved):

zakird:
I don't have strong opinions, I don't actually know the use case when you'd want this to be different, so I think take the path of least resistance and we can revisit if need be.

PR #448, opened 2024-09-19
Lowers default thread count from 1000 to 100

phillip-stephens:
After some investigation on thread count vs. runtime/failure rates, IMO there's sufficient evidence that 1k threads is too high.

Lowering this default to `threads=100` and users can further tune if they require more performance with the risk of getting a higher failure rate.

image


zakird:
It's hard to tell what "too high" is because with retries, 1000 is probably much faster with the same results than 100 with no retries. That said I'm not opposed to defaulting to 100 and leaving it to the user to go higher. Running 4x faster with 1% more failures isn't the worse tradeoff in many cases when you can just retry that 1%.

zakird (approved):

PR #456, opened 2024-09-27
Prevent NONEEDEDGLUE errors and add cache statistics

phillip-stephens:
## Description
Prior to this change, we were adding `Authorities` and `Additionals` as separate records in the cache. This meant that the `Authorities` was inserted as an NS record, but the Additionals were inserted as separate A/AAAA records. Separately, our cache is an LRU cache with 4096 shards, 2 entries per shard by default. What this means is that we have no guarantees that all of a domain's Authorities + Additionals are in the cache, since a partial set can be evicted.

This loss of accuracy was what caused the `NONEEDEDGLUE` records to pop up, especially as more domains in the same run.

## Changes
- Add CacheStats that capture cache hits, misses, evictions with `--verbosity=5`. No performance impact if `--verbosity < 5`
   - Ex: `time="2024-09-27T20:28:52Z" level=debug msg="Cache statistics: hits=29870 misses=85581 adds=30790 ejects=21604 hitRate=25.872448% missRate=74.127552%"`
- Switched to using a single `SingleQueryResult` pointer vs. returning the struct since each function would have to create a new structure and copy the internals. The pointer avoids this.
- Re-wrote cache so that an entire `Authority` record or `Answer record are stored atomically
    - `Authority` records store the authority's for a certain domain - `.com`
    - `Answer` records store the authoritative answers for a certain query - `A google.com`

## Performance
Tested with 10000 top domains, `A --iterative --threads=100 --verbosity=5`
### main
Runtime - 37.2 s
Allocs - 2199 MB total allocations (this is not peak memory use since the Garbage Collector cleans up un-used memory, this is just total allocated objects)
NOERROR - 9919
TIMEOUT/ITERATIVE_TIMEOUT - 5
NXDOMAIN - 54

### Branch
Runtime - 33.9 s
Allocs - 1216 MB
NOERROR - 9916
TIMEOUT/ITERATIVE_TIMEOUT - 9
NXDOMAIN - 55

Main takeaway is there was a sharp reduction in allocations, however the runtime reduction wasn't that large (and likely within the margin of error, other A/B tests didn't show such a large delta), so it seems garbage collection is not our bottleneck.

Either way, I think this is a needed change for cache use.

### DIFF
--- a/src/cli/worker_manager.go
+++ b/src/cli/worker_manager.go
@@ -530,6 +530,8 @@ func Run(gc CLIConf) {
 	close(outChan)
 	close(metaChan)
 	routineWG.Wait()
+	// print cache stats
+	//printCacheStats(resolverConfig.Cache)
 	if gc.MetadataFilePath != "" {
 		// we're done processing data. aggregate all the data from individual routines
 		metaData := aggregateMetadata(metaChan)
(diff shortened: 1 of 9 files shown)


src/cli/worker_manager.go
```diff
@@ -530,6 +533,7 @@ func Run(gc CLIConf) {
 	close(outChan)
 	close(metaChan)
 	routineWG.Wait()
+	resolverConfig.Cache.Stats.PrintStatistics()
```
zakird: We should move into metadata

zakird (approved):

PR #465, opened 2024-10-22
Refactor: Allow More Answer Types to Be Cacheable

developStorm:
This PR introduces the `WithBaseAnswer` interface as the cache value type to allow broader answer types to be included in the cache. The `answers_generate.go` script, when run with `go generate`, implements the required interface function `BaseAns` for each eligible answer struct.


src/zdns/answers_generate.go
```diff
@@ -0,0 +1,102 @@
+//go:build ignore
```
phillip-stephens, line 1: Let's add the license and attribution linking to the miekg project.

src/zdns/cache.go
```diff
@@ -169,68 +169,68 @@ func (s *Cache) getCachedResult(q Question, ns *NameServer, isAuthority bool, de
...
 }
 
 func (s *Cache) buildCachedResult(res *SingleQueryResult, depth int, layer string) *CachedResult {
 	now := time.Now()
 	cachedRes := CachedResult{}
 	cachedRes.Answers = make([]TimedAnswer, 0, len(res.Answers))
-	for _, a := range res.Answers {
-		castAns, ok := a.(Answer)
+
+	var getExpirationForSafeAnswer = func(a any) (WithBaseAnswer, time.Time) {
```
phillip-stephens, line 191: I like this helper fn, cleans up things towards the end of the function!

phillip-stephens (commented):
LGTM once we add the license stuff

phillip-stephens:
And for context, this is going to enable us to cache `DNSKEY` and `DS` entries so @developStorm can add DNSSEC validation support into ZDNS.

src/zdns/answers_generate.go
developStorm: added in https://github.com/zmap/zdns/pull/465/commits/6f09067a5b2b0e87abf4163ec2c12709f85bb37f

dadrian (approved):

PR #468, opened 2024-10-31
refactor: use replace syntax to import zdns/dns

developStorm:
This change enables data type compatibility for interoperability with libraries that reference `miekg` types. Currently, it allows us to import [go-dns-root-anchors](https://github.com/zmap/go-dns-root-anchors), but it also provides future extensibility. If we introduce other dependencies that rely on `miekg/dns`, this adjustment will enable us to transparently replace `miekg/dns` with our version.

Additionally, it ensures that `zmap/dns` interfaces and data types remain aligned with `miekg/dns`, guarding for inadvertent compatibility issues.

**Alternatives considered:**

- Requiring `zmap/dns` for `go-dns-root-anchors`: While this approach works, it would limit the utility of the root anchors library to primarily our use case. Additionally, if we need to add another dependency that relies on `miekg/dns` and isn’t under our control, we’d encounter the same compatibility issue.


developStorm:
~~Pending on https://github.com/golang/go/issues/70137~~

Workaround: https://github.com/zmap/dns/pull/14

go.mod
```diff
@@ -5,18 +5,20 @@ go 1.20
...
 	github.com/stretchr/testify v1.9.0
-	github.com/zmap/dns v1.1.59-zdns-4
 	github.com/zmap/go-iptree v0.0.0-20210731043055-d4e632617837
 	github.com/zmap/zcrypto v0.0.0-20240803002437-3a861682ac77
 	github.com/zmap/zflags v1.4.0-beta.1.0.20200204220219-9d95409821b6
 	github.com/zmap/zgrab2 v0.1.8
 	gotest.tools/v3 v3.5.1
 )
 
+replace github.com/miekg/dns => github.com/zmap/dns v1.1.63-zdns1
```
developStorm, line 20: `zmap/dns v1.1.63*` is `miekg/dns v1.1.62`. New tags are just to bypass go proxy cache system.

zakird (approved):

PR #470, opened 2024-11-01
feat: DNSSEC Validation

### LINKED ISSUE #441: Add DNSSEC Validation
**Describe the feature requested**
Add DNSSEC record validation to ZDNS. We currently have the `--dnssec` CLI flag to request DNSSEC `RRSIG` records. The ask here is to verify these signatures going up to the root of trust.

This feature should:
- Add a new `--validate-dnssec` CLI flag
- Make use of the `Resolver.cache` to avoid duplicating lookups
- Add integration tests to `integration_tests.py` for DNSSEC validation
- At the moment, we are only concerned with existence cases; i.e. NSEC(3) is out of scope.

Test Cases
- Positive Test Cases (from [internetsociety.org](https://www.internetsociety.org/resources/deploy360/2013/dnssec-test-sites/)
    - Cloudflare.com
    - internetsociety.org
    - dnssec-tools.org
    - dnssec-deployment.org
- Negative Test Cases
    - dnssec-failed.org
    - rhybar.cz
- `./zdns A dnssec-tools.org internetsociety.org  --validate-dnssec`
    - Check Wireshark for 0 duplicate DNS queries

Output
WIP - open to suggestions
Currently thinking we return the same information as `./zdns --dnssec` but 2 additional per module fields
- `dnssec-validation-passed: true/false`
- `dnssec-validation-failed-reason: "Signature of Cloudflare.com did not validate using the .com signing key"`

developStorm:
resolves #441

# Purpose
Adds support for validating DNSSEC records up to the root's chain-of-trust

The level of detail in the validation results depends on the `--result-verbosity=` flag. In the `short` mode, validation results are not output (equivalent to no validation). In the `normal` mode, only the overall validation status is output. In the `long` mode, all relevant records used for validating this specific result (DS, DNSKEY, RRSIG) and the validation results for each individual RRset are printed. In the `trace` mode, all validation details for the entire query chain are available along the trace.

Since we are a DNS scanning tool, the resolution results may still hold value even in cases of DNSSEC validation failure. Therefore, a validation failure does not lead to a resolution failure - i.e., the resolution results will still be output normally. It is the responsibility of data consumers to check the DNSSEC validation status before using the results.

Example usage:
```shell
go run . A example.com --iterative --validate-dnssec --result-verbosity=long
```

Example output (long):
```json
{
    "class": "IN",
    "name": "example.com",
    "results": {
        "A": {
            "data": {
                "answers": [
                    {
                        "answer": "93.184.215.14",
                        "class": "IN",
                        "name": "example.com",
                        "ttl": 3600,
                        "type": "A"
                    },
                    {
                        "algorithm": 13,
                        "class": "IN",
                        "expiration": "20241206041826",
                        "inception": "20241115121713",
                        "keytag": 60915,
                        "labels": 2,
                        "name": "example.com",
                        "original_ttl": 3600,
                        "signature": "vRP5fmIUOIHdIXJsnFkSLYL84OEh4isvOeIkPdGR1Ro+YUs4tTNO6J3ajnDvcIZfc5p+j8R35Z/SOOqD9ZY95w==",
                        "signer_name": "example.com.",
                        "ttl": 3600,
                        "type": "RRSIG",
                        "type_covered": 1
                    }
                ],
                "dnssec": {
                    "additionals": [
                        {
                            "error": "",
                            "rrset": {
                                "class": 4096,
                                "name": ".",
                                "type": 41
                            },
                            "sig": null,
                            "status": "Insecure"
                        }
                    ],
                    "answer": [
                        {
                            "error": "",
                            "rrset": {
                                "class": 1,
                                "name": "example.com.",
                                "type": 1
                            },
                            "sig": {
                                "algorithm": 13,
                                "class": "IN",
                                "expiration": "20241206041826",
                                "inception": "20241115121713",
                                "keytag": 60915,
                                "labels": 2,
                                "name": "example.com",
                                "original_ttl": 3600,
                                "signature": "vRP5fmIUOIHdIXJsnFkSLYL84OEh4isvOeIkPdGR1Ro+YUs4tTNO6J3ajnDvcIZfc5p+j8R35Z/SOOqD9ZY95w==",
                                "signer_name": "example.com.",
                                "ttl": 3600,
                                "type": "RRSIG",
                                "type_covered": 1
                            },
                            "status": "Secure"
                        }
                    ],
                    "authoritative": [
                        {
                            "error": "",
                            "rrset": {
                                "class": 1,
                                "name": "example.com.",
                                "type": 2
                            },
                            "sig": {
                                "algorithm": 13,
                                "class": "IN",
                                "expiration": "20241206020003",
                                "inception": "20241115121713",
                                "keytag": 60915,
                                "labels": 2,
                                "name": "example.com",
                                "original_ttl": 86400,
                                "signature": "9W8bAJT/4vA3nH8QU/NeM1n8bZZWn5PBkN26ix827VQGWY0YdMniks/1YCIaVNl16xly4s5IsVDtI9rLvaZORw==",
                                "signer_name": "example.com.",
                                "ttl": 86400,
                                "type": "RRSIG",
                                "type_covered": 2
                            },
                            "status": "Secure"
                        }
                    ],
                    "dnskey": [
                        {
                            "algorithm": 13,
                            "class": "IN",
                            "flags": 256,
                            "name": "example.com",
                            "protocol": 3,
                            "public_key": "ai2pvpijJjeNTpBu4yg6T375JqIStPtLABDTAILb+f4J7XpofUNXGQn6FpQvZ6CARWn2xQapbjGtDRjTf4qYxg==",
                            "ttl": 3600,
                            "type": "DNSKEY"
                        }
                    ],
                    "ds": [
                        {
                            "algorithm": 13,
                            "class": "IN",
                            "digest": "be74359954660069d5c63d200c39f5603827d7dd02b56f120ee9f3a86764247c",
                            "digest_type": 2,
                            "key_tag": 370,
                            "name": "example.com",
                            "ttl": 3600,
                            "type": "DS"
                        }
                    ],
                    "status": "Secure"
                },
                "flags": {
                    "authenticated": false,
                    "authoritative": true,
                    "checking_disabled": false,
                    "error_code": 0,
                    "opcode": 0,
                    "recursion_available": false,
                    "recursion_desired": false,
                    "response": true,
                    "truncated": false
                },
                "protocol": "udp",
            },
            "duration": 1.072630625,
            "status": "NOERROR",
        }
    }
}
```

Example output (normal):
```json
{
    "name": "example.com",
    "results": {
        "A": {
            "data": {
                "answers": [
                    {
                        "answer": "93.184.215.14",
                        "class": "IN",
                        "name": "example.com",
                        "ttl": 3600,
                        "type": "A"
                    },
                    {
                        "algorithm": 13,
                        "class": "IN",
                        "expiration": "20241206041826",
                        "inception": "20241115121713",
                        "keytag": 60915,
                        "labels": 2,
                        "name": "example.com",
                        "original_ttl": 3600,
                        "signature": "vRP5fmIUOIHdIXJsnFkSLYL84OEh4isvOeIkPdGR1Ro+YUs4tTNO6J3ajnDvcIZfc5p+j8R35Z/SOOqD9ZY95w==",
                        "signer_name": "example.com.",
                        "ttl": 3600,
                        "type": "RRSIG",
                        "type_covered": 1
                    }
                ],
                "dnssec": {
                    "status": "Secure"
                },
                "protocol": "udp"
            },
            "duration": 0.803631958,
            "status": "NOERROR"
        }
    }
}
```

- [x] Implement error status handling according to [RFC 4035, Section 4.3](https://www.rfc-editor.org/rfc/rfc4035.html)
- [x] Ensure proper handling of non-secure cases
- [x] Share retries 
- [x] Address all linter issues
- [x] Limit DNSSEC to iterative mode only?
- [x] CLI option

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -91,7 +91,7 @@ type InputOutputOptions struct {
 	BlacklistFilePath            string `long:"blacklist-file" description:"blacklist file for servers to exclude from lookups"`
 	DNSConfigFilePath            string `long:"conf-file" default:"/etc/resolv.conf" description:"config file for DNS servers"`
 	MultipleModuleConfigFilePath string `short:"c" long:"multi-config-file" description:"config file path for multiple module"`
-	IncludeInOutput              string `long:"include-fields" description:"Comma separated list of fields to additionally output beyond result verbosity. Options: class, protocol, ttl, resolver, flags"`
+	IncludeInOutput              string `long:"include-fields" description:"Comma separated list of fields to additionally output beyond result verbosity. Options: class, protocol, ttl, resolver, flags, dnssec"`
 	InputFilePath                string `short:"f" long:"input-file" default:"-" description:"names to read, defaults to stdin"`
 	LogFilePath                  string `long:"log-file" default:"-" description:"where should JSON logs be saved, defaults to stderr"`
 	MetadataFilePath             string `long:"metadata-file" description:"where should JSON metadata be saved, defaults to no metadata output. Use '-' for stderr."`
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -64,6 +64,7 @@ type QueryOptions struct {
 	ClassString        string `long:"class" default:"INET" description:"DNS class to query. Options: INET, CSNET, CHAOS, HESIOD, NONE, ANY."`
 	ClientSubnetString string `long:"client-subnet" description:"Client subnet in CIDR format for EDNS0."`
 	Dnssec             bool   `long:"dnssec" description:"Requests DNSSEC records by setting the DNSSEC OK (DO) bit"`
+	ValidateDNSSEC     bool   `long:"validate-dnssec" description:"Validate DNSSEC records, only applicable with --iterative"`
 	UseNSID            bool   `long:"nsid" description:"Request NSID."`
 }
 
--- a/src/zdns/conf.go
+++ b/src/zdns/conf.go
@@ -49,7 +49,16 @@ const (
 	StatusIterTimeout  Status = "ITERATIVE_TIMEOUT"
 	StatusNoAuth       Status = "NOAUTH"
 	StatusNoNeededGlue Status = "NONEEDEDGLUE" // When a nameserver is authoritative for itself and the parent nameserver doesn't provide the glue to look it up
+	StatusCircular     Status = "CIRCULAR"     // When circular query dependencies are detected
 )
+
+func isStatusRetryable(status Status) bool {
+	switch status {
+	case StatusServFail, StatusNXDomain, StatusRefused, StatusTruncated, StatusError, StatusTimeout, StatusIterTimeout:
+		return true
+	}
+	return false
+}
 
 var RootServersV4 = []NameServer{
 	{IP: net.ParseIP("198.41.0.4"), Port: 53},     // A
--- a/src/zdns/dnssec_types.go
+++ b/src/zdns/dnssec_types.go
@@ -142,8 +142,10 @@ func (r *DNSSECResult) populateStatus() {
 
 	for _, result := range r.Answer {
 		if result.Status == DNSSECInsecure {
-			r.Status = DNSSECInsecure
-			r.Reason = result.Error
+			// This is considered bogus. If we are at this point, we know a DS exists for
+			// the zone, so the answer section (authoritative data) should be signed.
+			r.Status = DNSSECBogus
+			r.Reason = "answer section is not signed when expected to be"
 			return
 		}
 
@@ -158,8 +160,8 @@ func (r *DNSSECResult) populateStatus() {
 		for _, result := range section {
 			if isDNSSECType(result.RRset.Type) {
 				if result.Status == DNSSECInsecure {
-					r.Status = DNSSECInsecure
-					r.Reason = result.Error
+					r.Status = DNSSECBogus
+					r.Reason = "DNSSEC-related RRset is not signed when expected to be"
 					return
 				}
 
--- a/src/zdns/dnssec_types.go
+++ b/src/zdns/dnssec_types.go
@@ -29,9 +29,19 @@ const (
 	DNSSECIndeterminate DNSSECStatus = "Indeterminate"
 )
 
+type RRsetKey struct {
+	Name  string `json:"name"`
+	Type  uint16 `json:"type"`
+	Class uint16 `json:"class"`
+}
+
+func (r *RRsetKey) String() string {
+	return "name: " + r.Name + ", type: " + dns.TypeToString[r.Type] + ", class: " + dns.ClassToString[r.Class]
+}
+
 // DNSSECPerSetResult represents the validation result for an RRSet
 type DNSSECPerSetResult struct {
-	RrType    string       `json:"rr_type"`
+	RRset     RRsetKey     `json:"rrset"`
 	Status    DNSSECStatus `json:"status"`
 	Signature *RRSIGAnswer `json:"sig"`
 	Error     string       `json:"error"`
@@ -91,9 +101,9 @@ func makeDNSSECResult() *DNSSECResult {
 // Otherwise, the overall status is secure.
 // This function should be called after all PerSetResults are populated, and the result should is stored in r.Status.
 func (r *DNSSECResult) populateStatus() {
-	isDNSSECType := func(rrType string) bool {
+	isDNSSECType := func(rrType uint16) bool {
 		switch rrType {
-		case "DNSKEY", "RRSIG", "DS", "NSEC", "NSEC3", "NSEC3PARAM":
+		case dns.TypeDNSKEY, dns.TypeRRSIG, dns.TypeDS, dns.TypeNSEC, dns.TypeNSEC3, dns.TypeNSEC3PARAM:
 			return true
 		default:
 			return false
@@ -127,7 +137,7 @@ func (r *DNSSECResult) populateStatus() {
 	// Check DNSSEC-related RRsets in other sections
 	for _, section := range [][]DNSSECPerSetResult{r.Additionals, r.Authoritative} {
 		for _, result := range section {
-			if isDNSSECType(result.RrType) {
+			if isDNSSECType(result.RRset.Type) {
 				if result.Status == DNSSECInsecure {
 					r.Status = DNSSECInsecure
 					return
--- a/src/zdns/dnssec_types.go
+++ b/src/zdns/dnssec_types.go
@@ -64,8 +64,8 @@ type dNSSECValidator struct {
 	nameServer  *NameServer
 	isIterative bool
 
-	ds     []*dns.DS
-	dNSKEY []*dns.DNSKEY
+	ds     map[dns.DS]bool
+	dNSKEY map[dns.DNSKEY]bool
 }
 
 // makeDNSSECValidator creates a new DNSSECValidator instance
@@ -77,8 +77,8 @@ func makeDNSSECValidator(r *Resolver, ctx context.Context, msg *dns.Msg, nameSer
 		nameServer:  nameServer,
 		isIterative: isIterative,
 
-		ds:     make([]*dns.DS, 0),
-		dNSKEY: make([]*dns.DNSKEY, 0),
+		ds:     make(map[dns.DS]bool),
+		dNSKEY: make(map[dns.DNSKEY]bool),
 	}
 }
 
--- a/src/zdns/dnssec_types.go
+++ b/src/zdns/dnssec_types.go
@@ -25,6 +25,7 @@ const (
 
 // DNSSECPerSetResult represents the validation result for an RRSet
 type DNSSECPerSetResult struct {
+	RrType       string
 	Status       DNSSECStatus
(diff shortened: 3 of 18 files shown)


phillip-stephens:
Additionally, let's add integration tests to `integration_tests.py` for positive/negative test cases, as defined here
https://www.internetsociety.org/resources/deploy360/2013/dnssec-test-sites/

src/zdns/resolver.go
```diff
@@ -269,8 +269,9 @@ type Resolver struct {
 	connInfoIPv4Loopback        *ConnectionInfo // used for IPv4 lookups to loopback nameservers
 	connInfoIPv6Loopback        *ConnectionInfo // used for IPv6 lookups to loopback nameservers
 
-	retries  int
-	logLevel log.Level
+	retries          int
+	retriesRemaining int
```
phillip-stephens: Can we add comments that `retries` is the static constant configured on resolver init and `retriesRemaining` is set/decremented for each domain's resolution to make the purpose clear

developStorm: 3561c36d1bcf1b0ea914cd31f2777e38b8d362a3

src/zdns/conf.go
```diff
@@ -49,8 +49,17 @@ const (
 	StatusIterTimeout  Status = "ITERATIVE_TIMEOUT"
 	StatusNoAuth       Status = "NOAUTH"
 	StatusNoNeededGlue Status = "NONEEDEDGLUE" // When a nameserver is authoritative for itself and the parent nameserver doesn't provide the glue to look it up
+	StatusCircular     Status = "CIRCULAR"     // When circular query dependencies are detected
 )
 
+func isStatusRetryable(status Status) bool {
+	switch status {
+	case StatusServFail, StatusNXDomain, StatusRefused, StatusTruncated, StatusError, StatusTimeout, StatusIterTimeout:
```
developStorm, line 57: @phillip-stephens could you check if these make sense to you

zakird:
What have we done to test this? Can we run this against the top million websites or such and see how it goes? A lot of code.

developStorm:
> What have we done to test this? 

At this point it's just manual and unit tests performed against the above domains shared by Phillip, and I monitored request sanity with Wireshark. I can look into large scale test later.

src/zdns/conf.go
phillip-stephens: Since `retries` are global to the lookup, we shouldn't use a `retry` if there's no chance of getting back a different response.

`NXDomain` we should not retry. Based on [RFC 8020](https://datatracker.ietf.org/doc/html/rfc8020), there's no chance we re-query and get a different response to an NXDomain. So this **should not** be retry-able
The others I'm fine with leaving as-is.

Additionally, let's add `NXDomain` as a specific `case` statement and add that RFC as reason why we will not be re-trying that specific domain for documentation.

And make `StatusTimeout` and `StatusIterTimeout` their own case statement and comment:
`up to caller's implementation to retry` since technically where this function is used in lookup.go, we'll abort the query if the `ctx` expires, the main reason for getting this error.

The others I'm fine with!

src/cli/cli.go
```diff
@@ -67,6 +67,7 @@ type QueryOptions struct {
 	ClassString        string `long:"class" default:"INET" description:"DNS class to query. Options: INET, CSNET, CHAOS, HESIOD, NONE, ANY."`
 	ClientSubnetString string `long:"client-subnet" description:"Client subnet in CIDR format for EDNS0."`
 	Dnssec             bool   `long:"dnssec" description:"Requests DNSSEC records by setting the DNSSEC OK (DO) bit"`
+	ValidateDNSSEC     bool   `long:"validate-dnssec" description:"Validate DNSSEC records, only applicable with --iterative"`
```
zakird, line 70: I think we discussed allowing this even when not in `--iterative``.

src/zdns/dnssec_types.go
```diff
@@ -0,0 +1,160 @@
...
+	Error     string       `json:"error"`
+}
+
+// DNSSECResult captures all information generated during a DNSSEC validation
+type DNSSECResult struct {
+	Status        DNSSECStatus         `json:"status" groups:"dnssec,dnssec,normal,long,trace"`
+	DS            []*DSAnswer          `json:"ds" groups:"dnssec,long,trace"`
+	DNSKEY        []*DNSKEYAnswer      `json:"dnskey" groups:"dnssec,long,trace"`
+	Answer        []DNSSECPerSetResult `json:"answer" groups:"dnssec,long,trace"`
+	Additionals   []DNSSECPerSetResult `json:"additionals" groups:"dnssec,long,trace"`
```
zakird: I think this should be `Additional`  not `Additionals` per RFC and in matching the other types.

src/zdns/resolver.go
```diff
@@ -88,15 +89,16 @@ type ResolverConfig struct {
...
-	VerifyServerCert    bool           // Verify server certificates for DoT/DoH
-	HTTPSClientIPv4     *http.Client   // for DoH, per docs should be shared amongst requests
-	HTTPSClientIPv6     *http.Client   // for DoH, per docs should be shared amongst requests
-	EdnsOptions         []dns.EDNS0
-	CheckingDisabledBit bool
+	DNSSecEnabled        bool
+	ShouldValidateDNSSEC bool           // whether to validate DNSSEC
+	DNSOverHTTPS         bool           // whether to use DNS over HTTPS for External Lookups, n/a to Iterative Lookups
+	DNSOverTLS           bool           // whether to use DNS over TLS for External Lookups, n/a to Iterative Lookups
+	RootCAs              *x509.CertPool // Root CAs for DoT/DoH Server Verification
```
zakird, line 97: Do we want to allow people to pass in root store on the command line? Otherwise, where are we expecting it to originate?

developStorm: DNSSEC root anchors are from https://github.com/zmap/go-dns-root-anchors. That repository updates itself as described in RFC7958.

src/zdns/dnssec_types.go
developStorm: We currently have `json:"additionals` in `SingleQueryResult` 
https://github.com/zmap/zdns/blob/d403578a6ca4aaed047ab6bc452e67b7df4a9419/src/zdns/qa.go#L77-L79

and in `CachedResult`:
https://github.com/zmap/zdns/blob/d403578a6ca4aaed047ab6bc452e67b7df4a9419/src/zdns/cache.go#L35-L38

I can address the pluralization issue for DNSSEC and cache (ideally in a separate PR). However, for `SingleQueryResult`, since the plural form is embedded in the JSON schema, changing it could introduce backward compatibility issues. 

If we leave it unchanged, we’ll have a mismatch where the JSON schema for DNSSEC uses the singular form, but the response retains the plural form. This inconsistency might be confusing.

To resolve this, I suggest we standardize by either:
1. Removing all plural forms and break compatibility, or 
2. Consistently using plural forms, at least for the JSON output.

What do you think?

zakird: We're about to do a major 2.0 release. I think we should have it consistent everywhere, even if it means that schema changes.

developStorm: Fixed in dnssec code. Created #484 for other places.

developStorm:
I conducted another round of tests on the top 10k domains, and the results aligned closely with Cloudflare, with a few explainable non-alignments remain:

1. **Domains under the `.my` TLD**:  
   These cannot be verified by us because they use an oversized RSA exponent that Go does not support. This issue is documented here: https://github.com/golang/go/issues/3161. As discussed with @phillip-stephens, we agree that per RFC 4033, this should be classified as bogus.

2. **Specific domains (`www.correoargentino.com.ar`, `xskt.com.vn`, and `futbollibrehd.pe`)**:  
   Cloudflare flagged these as bogus due to their inability to establish Secure Entry Point (SEP). Upon reviewing [DNSViz](https://dnsviz.net/d/xskt.com.vn/dnssec/), it appears these domains use NSEC3 records with iterations > 0, which is non-compliant with [RFC9276](https://datatracker.ietf.org/doc/html/rfc9276#section-3.1). Cloudflare likely ignored these NSEC3 records, rendering them unable to confirm the non-existence of DS records.  

   Since we are not a publicly facing recursive resolver and are not exposed to the risk of DoS attacks from this, I decided that we would continue to validate these NSEC3 records while printing a log: 9f7d8913ef1e5af17151c80f74d2c26ab4034452.

I'm comfortable with the shape of this PR now. Feel free to start reviewing @phillip-stephens @zakird.

src/cli/cli.go
developStorm: DM'ed on this and explained why it's not feasible. If we reach a different conclusion later, happy to implement this in a separate PR.

zakird (approved):
Overall this look great and I'm very excited to see DNSSEC validation added. @phillip-stephens is going to take a closer look.

zakird:
Re 1: This might be worth fixing in ZCrypto and using that. I think that we can leave this however as a FR and merge in any case.

phillip-stephens (approved):
LGTM, I tested it locally on 1k domains and spot-checked a few of them. Also reverted the change to require `go 1.20` since not sure we need to do that. Otherwise this looks really great and super excited about it! Thanks @developStorm!

PR #477, opened 2024-11-19
feat: lazy query A/AAAA of NSes when iterating on authorities

developStorm:


src/zdns/lookup.go
```diff
@@ -995,7 +996,12 @@ func (r *Resolver) extractAuthority(ctx context.Context, authority interface{},
 			q.Q.Type = dns.TypeA
 		}
 		q.RetriesRemaining = &r.retriesRemaining
+
+		// A/AAAA records for NSes are not on the chain of trust, so we don't need to validate DNSSEC
+		// Doing this to save us some time (this can propogate A LOT of queries in certain cases)
+		r.shouldValidateDNSSEC = false
 		res, trace, status, _ = r.iterativeLookup(ctx, &q, r.rootNameServers, depth+1, ".", trace)
+		r.shouldValidateDNSSEC = true
```
phillip-stephens: Do we want to re-enable this in all cases? What if we're not validating DNSSEC?

src/zdns/lookup.go
```diff
@@ -897,30 +897,35 @@ func (r *Resolver) iterateOnAuthorities(ctx context.Context, qWithMeta *Question
 	if len(result.Authorities) == 0 {
```
phillip-stephens, line 899: Let's comment `iterateOnAuthorities` (know it didn't have one before you, we'll just explain the behavior here)
Something like:

// iterateOnAuthorities takes the authorities from the referrals of a nameserver, shuffles them, and iteratively tries to do a lookup against them. If one succeeds, we return without trying the others. If one fails, we iterate to the next.

phillip-stephens (commented):
Few suggestions, otherwise looks good to me.

Since we're changing the lookup behavior for all lookups, can you add a `TODO` to the main DNSSEC PR to do A/B tests with DNSSEC disabled to be sure our error rate/timeout rate/success rate/runtime are all similar or better with that branch than `main` ZDNS? Want to check for regressions.

src/zdns/lookup.go
developStorm: Good catch! 989055717374f7f39a108b4cbad318596ed4acb9

developStorm:
Merging this now to the DNSSEC dev branch.

developStorm:
For A/B testing, current main we have
```
01h:02m:25s; Scan Complete, no more input. 1000000 domains scanned; 266.98 domains/sec.; 99.0% success rate; NOERROR: 984428, NXDOMAIN: 5321, REFUSED: 5094, TIMEOUT: 3160, ITERATIVE_TIMEOUT: 1075, SERVFAIL: 477, ERROR: 392, FORMERR: 32, AUTHFAIL: 12, NOTAUTH: 7, NOTIMP: 2
```

With this change (and DNSSEC on, and with Cloudflare DoH):
```
01h:43m:19s; Scan Complete, no more input. 984002 domains scanned; 158.72 domains/sec.; 99.3% success rate; NOERROR: 971703, SERVFAIL: 5355, NXDOMAIN: 5342, TIMEOUT: 1602
```

From error rate I think we are good? Current main might get penalized for scanning too fast I assume.

PR #487, opened 2024-12-18
Unify instances of "additionals" -> "additional" when referring to the section as a whole

### LINKED ISSUE #484: Inconsistent Pluralization for Additional Section
See https://github.com/zmap/zdns/pull/470#discussion_r1857223120. We should make it singular everywhere.

phillip-stephens:
Resolves #484 

[RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035#section-4.1) - refers to the `Additional` section in the singular. 

This PR fixes that so our output sections align with the RFC. I've left other instances of `additionals` in the code since it makes the code more readable if you refer to a set of items in the plural.


zakird (approved):

src/cli/modules.go
```diff
@@ -58,7 +58,7 @@ func init() {
 	RegisterLookupModule("CSYNC", &BasicLookupModule{DNSType: dns.TypeCSYNC, DNSClass: dns.ClassINET})
 	RegisterLookupModule("DHCID", &BasicLookupModule{DNSType: dns.TypeDHCID, DNSClass: dns.ClassINET})
 	RegisterLookupModule("DNAME", &BasicLookupModule{DNSType: dns.TypeDNAME, DNSClass: dns.ClassINET})
-	RegisterLookupModule("DNSKEY", &BasicLookupModule{DNSType: dns.TypeDNSKEY, DNSClass: dns.ClassINET})
+	RegisterLookupModule("DNSKEYs", &BasicLookupModule{DNSType: dns.TypeDNSKEY, DNSClass: dns.ClassINET})
```
zakird: This feels like if the Type is DNSKey we should leave it singular like all the others.

phillip-stephens: Good catch, I agree

zakird (approved):

PR #491, opened 2024-12-20
Phillip/483 ipv6 bug

### LINKED ISSUE #483: [Bug] V6 iterative A lookup of example.com returns SERVFAIL
**Describe the bug**

ZDNS fails to lookup the A record of example.com via IPv6, even though example.com is resolvable via V6 and name servers with AAAA are available.

**To Reproduce**
Version (HEAD)

> echo example.com. | ./zdns A --iterative --6
```
{"altered_name":"example.com","name":"example.com.","results":{"A":{"data":{"protocol":"","resolver":""},"duration":0.681134153,"error":"no valid nameservers found","status":"SERVFAIL","timestamp":"2024-12-04T09:50:35Z"}}}
00h:00m:00s; Scan Complete; 1 names scanned; 1.47 names/sec; 0.0% success rate; SERVFAIL: 1
``` 

> echo example.com. | ./zdns A --iterative --4
```
{"altered_name":"example.com","name":"example.com.","results":{"A":{"data":{"answers":[{"answer":"93.184.215.14","class":"IN","name":"example.com","ttl":3600,"type":"A"}],"protocol":"udp","resolver":"199.43.135.53:53"},"duration":0.911606947,"status":"NOERROR","timestamp":"2024-12-04T09:40:46Z"}}}
00h:00m:00s; Scan Complete; 1 names scanned; 1.10 names/sec; 100.0% success rate; NOERROR: 1
```

**Expected behavior**
Same output as for `--4` (modulo IP-addresses of the name servers)

**Additional context**
Looking at packet captures, there appears to be no attempt to lookup the AAAA records for `a.iana-servers.net` and `b.iana-servers.net` which are authoritative for `example.com`. The iterative lookup fails 

Potential Cause:
Should this be an "or" instead of "and"?

https://github.com/zmap/zdns/blob/main/src/zdns/lookup.go#L962
```
		if r.ipVersionMode != IPv4Only && r.iterationIPPreference == PreferIPv6 {
			q.Q.Type = dns.TypeAAAA
		} else {
			q.Q.Type = dns.TypeA
		}
```

phillip-stephens:
Resolves #483 

@steurer was right, we had a boolean logic error which partially fixed this. I noticed a few other logic issues during debugging. This PR fixes the issue on Ubuntu, Mac is a bit more temperamental, see below.

## Testing
On my Mac, I'm having issues even still with getting timeouts ~10% of the time when exchanging IPv6 DNS msgs on-the-wire. It's only to certain nameservers, and it's consistent ie. the same nameserver experiences the timeout.

Tested on an Ubuntu VM with IPv6 connectivity around x20 and didn't run into an issue:
`go run . A example.com --6 --iterative;`

I'm guessing there's some issue with IPv6 on either my local network or configuration.

```sh
root@ubuntu-s-1vcpu-2gb-nyc1-01:~/zdns# go run . A example.com --6 --iterative
WARN[0000] none of the user-supplied local addresses could connect to name server [2001:503:c27::2:30]:53, using local address 2604:a880:400:d1::9c4b:b001
00h:00m:00s; Scan Complete; 1 names scanned; 8.00 names/sec; 100.0% success rate; NOERROR: 1
{"name":"example.com","results":{"A":{"data":{"answers":[{"answer":"93.184.215.14","class":"IN","name":"example.com","ttl":3600,"type":"A"}],"protocol":"udp","resolver":"[2001:500:8d::53]:53"},"duration":0.125344131,"status":"NOERROR","timestamp":"2024-12-21T00:07:21Z"}}}
root@ubuntu-s-1vcpu-2gb-nyc1-01:~/zdns#
```


src/cli/worker_manager.go
```diff
@@ -235,6 +235,11 @@ func populateResolverConfig(gc *CLIConf) *zdns.ResolverConfig {
 			log.Fatal("unable to parse blacklist file: ", err)
 		}
 	}
+	// This must occur after setting the DNSConfigFilePath above, so that ZDNS knows where to fetch the DNS Config
+	config, err := populateIPTransportMode(gc, config)
+	if err != nil {
+		log.Fatal("could not populate IP transport mode: ", err)
+	}
 	// This is used in extractAuthorities where we need to know whether to request A or AAAA records to continue iteration
 	// Must be set after populating IPTransportMode
```
phillip-stephens, line 244: Noticed we said this had to be "after populating IPTransportMode" and set transport mode after 😢

zakird (approved):

PR #495, opened 2024-12-26
update golang to v1.23

stklcode:
golang 1.21 is EOL since August and 1.22 will follow around February. So let's jump to the latest supported version 1.23. (and run `go mod tidy`)

CI passes and quick manual tests don't show any regression.


zakird (approved):
Thanks — makes a lot of sense to me.

Seanstoppable:
We didn't seem to want to make this jump for zgrab2 in https://github.com/zmap/zgrab2/pull/466

Does zdns not have the same issues, or do we not particularly care to maintain consistency across the projects?

stklcode:
Today 1.22 is available pre-packaged for almost any semi-modern OS of the last 8 years or so. 1.23 is also okay, guess that was not the case back in September, few days after the release. And if I got it right, it’s only about compiling from source.

Stalin 1.21 already fails most security scanners today as well as statically linked binaries do (did not run any deeper analysis on the actual impact -  if any - just noticed the crit warning on my minimalistic 2.0.0-RC1 container build)

1.22 should not make much of a difference and could increase compatibility for projects including the code as a library - quite a fair point. Build process for binaries is a pretty weak argument though IMHO, this can be easily delegated to a build system, container or similar running a recent golang version.

zakird:
I think it's fine to bump Zgrab to 1.23 too.

PR #497, opened 2025-01-09
--help small format fix

phillip-stephens:
Since `zdns` is normally installed as an executable and not run from the directory, I wanted to refer to it here as `zdns` instead of `./zdns`


zakird (approved):

zakird:
```
Usage
Domains can be taken from a new-line delimited file, stdin, or passed as arguments.
./zdns A --input-file domains.txt
echo "google.com" | ./zdns A
cat domains.txt | ./zdns A
./zdns A google.com yahoo.com
Name Servers can either be specified as a global list or a per-domain name server.
./zdns A google.com --name-servers=1.1.1.1,8.8.8.8
echo "google.com,1.0.0.1" | ./zdns A
For Example:
zdns A google.com --name-servers=1.1.1.1,8.8.8.8
echo "google.com,1.0.0.1" | zdns A
```

I think choose one way or the other ;)

PR #498, opened 2025-01-09
Fix go-processes --help text

phillip-stephens:
Before: `      --go-processes=          number of OS processes (GOMAXPROCS by default) (default: 0)`
After: `--go-processes=          number of OS processes to use, GOMAXPROCS if 0 (default: 0)`


zakird (approved):
once tests pass ;)

zakird (approved):

PR #499, opened 2025-01-09
Add a Github Action to update the TLSA record in our integration tests automatically

phillip-stephens:
- Run python formatter `black` on our `integration_tests.py` (Reason for most changes)
- Add GitHub action to update TLSA record when `dig` shows it's changed
- Add `black` to our usual lint routine (both in `makefile` and in the lint `action` that runs on all PR's


phillip-stephens:
@zakird So every month or two the `TLSA` record we test against gets updated. This action should make a PR when that happens. However, I can't fully test the Github Action workflow until it's merged into `main`. Right now this just fails because the `TLSA` record is not correct. If it looks good to you, can we merge and then we can be sure the action creates a PR as intended?

zakird (approved):

zakird (approved):

PR #538, opened 2025-04-10
Implement full context support in ZDNS and better SIGTERM/SIGINT handling

phillip-stephens:
On Mac, there's always been an annoying behavior (existing back to at least `v1.1.0` that I saw) where SIGINT (`ctrl + C`) wouldn't kill ZDNS. I've had ZDNS continue after SIGINT for multiple minutes, and I end up just closing the terminal and opening another. Linux doesn't seem to have this behavior and ZDNS will exit immediately.

This PR implements both:

- Better handling of `context.Context`'s in ZDNS. Now the main API's (`ExternalLookup()`, etc.) all accept a `ctx` object.
- Better handling of SIGINT when using ZDNS as a CLI tool. Now, a goroutine will watch for `SIGINT` and `SIGTERM` and cancel the main `context.Context`. This ensures ZDNS exits quickly on Mac.

### 2 Second Delay in Cancellation
With this PR, ZDNS on Mac continues for 2-3 seconds post SIGINT. Using stack dumps, I noticed all goroutines are stuck in a `zmap/dns` lookup. At first this was really surprising since all the `dns` lookups we do use a context object which should be canceled when the main context is. However, looking closer at the [source code](https://github.com/zmap/dns/blob/master/client.go#L227) for `zmap/dns`,  only the `ctx.Deadline` is used to configure timeouts on the DNS connection object, and our default `--network-timeout` is `2s`.  So we can't cancel in-progress DNS requests and will have to let them expire. Regardless, this seems strictly better than our current state.

### Final Status Update
~~Currently, the status handler will print a final `Scan Complete; X names scanned...` with SIGINT. Wasn't sure if it'd be better to just immediately quit or print this.~~

Update - Per PR discussion, we now print a `Scan Aborted` instead of `Scan Complete` to indicate to the user we aborted due to their signal and didn't exit cleanly.

### v1.1.0 Behavior

https://github.com/user-attachments/assets/201e911c-6ee5-41fd-a9bf-fe92aad8f620

### This PR Behavior

https://github.com/user-attachments/assets/a2330365-76a3-4f44-bc8d-85d087872ac3


phillip-stephens:
@zakird Wondering if you have thoughts on the Final Status Update Limitation I listed in the description? Otherwise this is good to go.

zakird (approved):

zakird:
> Currently, the status handler will print a final Scan Complete; X names scanned... with SIGINT. Wasn't sure if it'd be better to just immediately quit or print this.
 
 I don't think it _super_ matters. If someone killed the process, they shouldn't really rely on the results. I'm fine leaving it has having it print the status message, but I thought that you might want to print something afterwards that indicates that the process has been interrupted so that the user doesn't inadvertently think that the process completed successfully.

Issue #539, opened 2025-04-13
[Bug] v2.0.3 Fresh Install Errors (tcpConn/udpConn Errors)


LloydLabs:
**Describe the bug**
Simply did `go get github.com/zmap/zdns/v2` (`v2.0.3`), using `go1.24.1`. I attempted to do a simple lookup:
```go
question := &zdns.Question{Name: domain, Type: dns.TypeA, Class: dns.ClassINET}
resp, _, _, _ := resolver.IterativeLookup(context.Background(), question)
```

Which resulted in on compilation:
```
../../go/pkg/mod/github.com/zmap/zdns/v2@v2.0.3/src/zdns/lookup.go:1148:98: connInfo.tcpConn.RemoteAddr.String undefined (type func() net.Addr has no field or method String)
../../go/pkg/mod/github.com/zmap/zdns/v2@v2.0.3/src/zdns/lookup.go:1156:34: connInfo.tcpClient.ExchangeWithConnToContext undefined (type *dns.Client has no field or method ExchangeWithConnToContext)
../../go/pkg/mod/github.com/zmap/zdns/v2@v2.0.3/src/zdns/lookup.go:1210:34: connInfo.udpClient.ExchangeWithConnToContext undefined (type *dns.Client has no field or method ExchangeWithConnToContext)
../../go/pkg/mod/github.com/zmap/zdns/v2@v2.0.3/src/zdns/resolver.go:524:99: connInfo.tcpConn.RemoteAddr.String undefined (type func() net.Addr has no field or method String)
../../go/pkg/mod/github.com/zmap/zdns/v2@v2.0.3/src/zdns/resolver.go:605:2: cannot assign to connInfo.tcpConn.RemoteAddr (neither addressable nor a map index expression)
```

Appears to be an issue with mainly `tcpConn`/`udpConn`.

Thanks,
Lloyd.

phillip-stephens:
Hey Lloyd!
Thanks for reaching out. So you'll need to do a few more setup steps before the `zdns.Resolver` will work. This [example](https://github.com/zmap/zdns/blob/main/examples/single_lookup/simple.go) has the full process (as well as you can run it directly), but you'll at least need to set the local IP and the root nameservers. If you'd like to use all Root nameservers for your `IterativeLookup` (the common case and what the CLI version of ZDNS defaults to), you'll want to set `config.RootNameServersV4 = zdns.RootNameServersV4`.

[...]

PR #542, opened 2025-04-15
Remove dependency on miekg/dns and only use the forked zmap/dns

phillip-stephens:
## Motivation
In issue #539, a user reported not being able to use the ZDNS library. After a bit of investigation, this is because in our repo we had added a line to the `go.mod` file:
```
replace github.com/miekg/dns => github.com/zmap/dns v1.1.65
```
This works fine for our repo, but `go get github.com/zmap/zdns` doesn't respect that line and so the library won't compile for users trying to import it.

## Description
The fix is to fully switch over to using `zmap/dns`, our fork of the dns library. This PR does that as well as adding `go.mod` files to the `examples/` folder.

## Implications
There was a bit of a chicken and egg issue with this. The `examples/` directory needs to import `zdns` but this fix isn't in a tagged version yet. So I have the examples importing a _commit-tagged_ version of ZDNS to get around this and so they will work out of the box. In `2.0.5` we can switch them to using the up-to-date tagged version.


LloydLabs:
Hey @phillip-stephens, awesome work! When is the next release planned with this fix introduced? As I plan to use this in my development cycle.

Best,
Lloyd.

zakird (approved):

phillip-stephens:
@LloydLabs I've now tagged a new `v2.0.4` with this fix in it. Let me know if you have any issues!
https://github.com/zmap/zdns/releases/tag/v2.0.4

PR #544, opened 2025-04-24
Parse LOC records into a user-readable format

### LINKED ISSUE #460: Add LOC record conversion into human-readable values
Currently we're just returning the raw DNS values for LOC record types.
Let's add parsing into human-readable values.
This would resolve the TODO [here](https://github.com/zmap/zdns/blob/main/src/zdns/answers.go#L736).

Ashleydai2003:
Helper function to convert raw DNS LOC values to human readable coordinates. The conversion follows RFC 1876 section 3 for LOC record format.

Unit Tests includes:
- All 4 hemisphere
- High altitude 
- Negative altitude 
- Greenwich (0)

## Related Issues
Resolves #460

### DIFF
--- a/src/zdns/answers.go
+++ b/src/zdns/answers.go
@@ -17,6 +17,7 @@ package zdns
 import (
 	"encoding/hex"
 	"fmt"
+	"math"
 	"net"
 	"strconv"
 	"strings"
@@ -144,6 +145,11 @@ type LOCAnswer struct {
 	Answer
 	Version     uint8  `json:"version" groups:"short,normal,long,trace"`
 	Size        uint8  `json:"size" groups:"short,normal,long,trace"`
+	HorizPre    uint8  `json:"horizontal_pre" groups:"short,normal,long,trace"`
+	VertPre     uint8  `json:"vertical_pre" groups:"short,normal,long,trace"`
+	Latitude    uint32 `json:"latitude" groups:"short,normal,long,trace"`
+	Longitude   uint32 `json:"longitude" groups:"short,normal,long,trace"`
+	Altitude    uint32 `json:"altitude" groups:"short,normal,long,trace"`
 	Coordinates string `json:"coordinates" groups:"short,normal,long,trace"`
 }
 
@@ -455,8 +461,8 @@ func euiToString(eui uint64, bits int) (hex string) {
 // The conversion follows RFC 1876 section 3 for LOC record format
 func formatLOCCoordinates(rawLat, rawLong, rawAlt uint32, size, horizPre, vertPre uint8) string {
 	// Convert raw ms values to signed integer values in seconds
-	latSeconds := float64((int64(rawLat) - 2147483648)) / 1000.0
-	longSeconds := float64((int64(rawLong) - 2147483648)) / 1000.0
+	latSeconds := float64((int64(rawLat) - math.MaxInt32 + 1)) / 1000.0
+	longSeconds := float64((int64(rawLong) - math.MaxInt32 + 1)) / 1000.0
 
 	// Determine hemispheres based on sign
 	latHemisphere := "N"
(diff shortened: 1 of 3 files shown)


zakird:
Hi! In this situation, I would add this as a new formatted coordinate string value and continue to provide the raw values since it'll be a pain to parse these out after the fact.

zakird (changes requested):
Hi! In this situation, I would add this as a new formatted coordinate string value and continue to provide the raw values since it'll be a pain to parse these out after the fact. Both could be useful in different situations. This will also make this a non-breaking change to the tool.

Also, I suspect already on your radar, but in case not, looks like tests are failing, which we'll need to get cleaned up.

src/zdns/answers.go
```diff
@@ -455,6 +451,50 @@ func euiToString(eui uint64, bits int) (hex string) {
 
 // <<<<< END GOOGLE CODE
 
+// formatLOCCoordinates converts raw DNS LOC values to human-readable GPS coordinates
+// The conversion follows RFC 1876 section 3 for LOC record format
+func formatLOCCoordinates(rawLat, rawLong, rawAlt uint32, size, horizPre, vertPre uint8) string {
+	// Convert raw ms values to signed integer values in seconds
+	latSeconds := float64((int64(rawLat) - 2147483648)) / 1000.0
```
phillip-stephens: Let's do `math.MaxInt32 + 1` so it's clear where `2147483648` came from and avoid magic numbers

src/zdns/answers_test.go
```diff
@@ -0,0 +1,124 @@
...
+ */
+package zdns
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFormatLOCCoordinates(t *testing.T) {
```
phillip-stephens, line 22: These look very comprehensive and clean, great work!

src/zdns/answers.go
```diff
@@ -455,6 +451,50 @@ func euiToString(eui uint64, bits int) (hex string) {
...
+	latDegrees := int32(latSeconds) / 3600
+	latMinutes := (int32(latSeconds) % 3600) / 60
+	latSeconds = latSeconds - float64(latDegrees*3600+latMinutes*60)
+
+	longDegrees := int32(longSeconds) / 3600
+	longMinutes := (int32(longSeconds) % 3600) / 60
+	longSeconds = longSeconds - float64(longDegrees*3600+longMinutes*60)
+
+	// Convert altitude from centimeters to meters
+	altMeters := (float64(rawAlt) / 100.0) - 100000.00
```
phillip-stephens: Let's add a comment of where the `... - 100000.00` comes from. Looking at the RFC, it seems to be part of the documented conversion, but as the code reads now, it's unclear why we need to do that. A short blurb of explanation and referring people to the RFC (with the section that discusses it) would suffice.

phillip-stephens (changes requested):
Thanks for the contribution @Ashleydai2003, this will be great for LOC usablility!

In addition to Zakir's concerns, I've added a few suggestions as well. Additionally, something is incorrect with the `precision` handling. With the example I setup on loc.prstephens.com, `dig` gives a different value for precision than this branch. Let's fix that discrepancy and add a unit test for a precision != 1m.

```
$ dig -t LOC loc.prstephens.com

; <<>> DiG 9.10.6 <<>> -t LOC loc.prstephens.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 60174
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;loc.prstephens.com.		IN	LOC

;; ANSWER SECTION:
loc.prstephens.com.	300	IN	LOC	37 25 26.000 N 122 9 57.000 W 30.00m 1m 1m 1m
```

```
$ ./zdns LOC loc.prstephens.com | jq ".results.LOC.data.answers[0].coordinates"
"37 25 26.000 N 122 9 57.000 W 30.00m 18m 18m 18m"
```

phillip-stephens (approved):
Looks good to me! Tested with a couple sample domains compared to `dig` and looks good. Thanks for this work!

zakird (approved):

PR #552, opened 2025-07-22
[WIP] added 32-bit integration test

phillip-stephens:


phillip-stephens:
Was attempting to have a 32-bit integration test, however it runs _really_ slow since it needs to emulate a 32-bit environment. With that in mind and the fact that this bug would have been present for months (or longer?)  and is incredibly easy to reproduce on a 32-bit system, this indicates to me this isn't a heavily used OS with ZDNS. We can always add it later if there is a demand. 

See [new PR](https://github.com/zmap/zdns/pull/553) with a cleaner commit history with just the fix, not integration test for it.

PR #559, opened 2025-09-04
Fix #548

### LINKED ISSUE #548: Support reverse addr lookups like dig -x
**Describe the feature requested**

```
 ./zdns PTR 8.8.8.8
00h:00m:00s; Scan Complete; 1 names scanned; 519.13 names/sec; 100.0% success rate; NOERROR: 1
{"name":"8.8.8.8","results":{"PTR":{"data":{"answers":[{"answer":"dns.google.","class":"IN","name":"8.8.8.8.in-addr.arpa","ttl":0,"type":"PTR"}],"protocol":"udp","resolver":"192.168.16.1:53"},"duration":0.001637673,"status":"NOERROR","timestamp":"2025-05-09T15:25:27+08:00"}}}

./zdns PTR 8.8.8.8.in-addr.arpa.
{"altered_name":"8.8.8.8.in-addr.arpa","name":"8.8.8.8.in-addr.arpa.","results":{"PTR":{"data":null,"duration":0.000218826,"error":"dns: unrecognized address: 8.8.8.8.in-addr.arpa","status":"ILLEGAL_INPUT","timestamp":"2025-05-09T15:19:09+08:00"}}}
00h:00m:00s; Scan Complete; 1 names scanned; 1085.56 names/sec; 0.0% success rate; ILLEGAL_INPUT: 1
```
`dig -x 8.8.8.8` and `dig 8.8.8.8.in-addr.arpa. PTR` has same behavior

I can also query some upper reverse addr NS record by `dig -x 8.8.8 NS`. Unfortunately, dns.ReverseAddr only support standard IPv4 or IPv6.

**Solution**
refer to bind9 dighost.c
```
isc_result_t
get_reverse(char *reverse, size_t len, char *value, bool strict) {
    int r;
    isc_result_t result;
    isc_netaddr_t addr;
    addr.family = AF_INET6;
    r = inet_pton(AF_INET6, value, &addr.type.in6);
    if (r > 0) {
        /* This is a valid IPv6 address. */
        dns_fixedname_t fname;
        dns_name_t *name;
        name = dns_fixedname_initname(&fname);
        result = dns_byaddr_createptrname(&addr, name);
        if (result != ISC_R_SUCCESS) {
            return result;
        }
        dns_name_format(name, reverse, (unsigned int)len);
        return ISC_R_SUCCESS;
    } else {
        /*
         * Not a valid IPv6 address.  Assume IPv4.
         * If 'strict' is not set, construct the
         * in-addr.arpa name by blindly reversing
         * octets whether or not they look like integers,
         * so that this can be used for RFC2317 names
         * and such.
         */
        char *p = reverse;
        char *end = reverse + len;
        if (strict && inet_pton(AF_INET, value, &addr.type.in) != 1) {
            return DNS_R_BADDOTTEDQUAD;
        }
        result = reverse_octets(value, &p, end);
        if (result != ISC_R_SUCCESS) {
            return result;
        }
        /* Append .in-addr.arpa. and a terminating NUL. */
        result = append(".in-addr.arpa.", 15, &p, end);
        if (result != ISC_R_SUCCESS) {
            return result;
        }
        return ISC_R_SUCCESS;
    }
}
```

zakird:
Fixes bug where we don't handle inputs like `8.8.8.8.in-addr.arpa` correctly in the PTR module. 

Resolves #548

### DIFF
--- a/src/zdns/lookup.go
+++ b/src/zdns/lookup.go
@@ -123,11 +123,9 @@ func (r *Resolver) doDstServersLookup(ctx context.Context, q Question, nameServe
 		qname, err = dns.ReverseAddr(q.Name)
 		// might be an actual DNS name instead of an IP address
 		// if that looks likely, use it as is
-		if err != nil {
-			if !util.IsStringValidDomainName(q.Name) {
-				return nil, nil, StatusIllegalInput, err
-			}
-		} else {
+		if err != nil && !util.IsStringValidDomainName(q.Name) {
+			return nil, nil, StatusIllegalInput, err
+		} else if len(qname) > 0 && qname[len(qname)-1] == '.' {
 			// remove trailing "." added by dns.ReverseAddr
 			q.Name = qname[:len(qname)-1]
 		}
--- a/src/zdns/lookup.go
+++ b/src/zdns/lookup.go
@@ -123,11 +123,11 @@ func (r *Resolver) doDstServersLookup(ctx context.Context, q Question, nameServe
 		qname, err = dns.ReverseAddr(q.Name)
 		// might be an actual DNS name instead of an IP address
 		// if that looks likely, use it as is
-		if err != nil && !util.IsStringValidDomainName(qname) {
-			return nil, nil, StatusIllegalInput, err
-			// q.Name is a valid name, we can continue
+		if err != nil {
+			if !util.IsStringValidDomainName(q.Name) {
+				return nil, nil, StatusIllegalInput, err
+			}
 		} else {
-			// remove trailing "." added by dns.ReverseAddr
 			q.Name = qname[:len(qname)-1]
 		}
 	}
--- a/src/zdns/lookup.go
+++ b/src/zdns/lookup.go
@@ -128,6 +128,7 @@ func (r *Resolver) doDstServersLookup(ctx context.Context, q Question, nameServe
 				return nil, nil, StatusIllegalInput, err
 			}
 		} else {
+			// remove trailing "." added by dns.ReverseAddr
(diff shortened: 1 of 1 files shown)


src/zdns/lookup.go
```diff
@@ -123,9 +123,10 @@ func (r *Resolver) doDstServersLookup(ctx context.Context, q Question, nameServe
 		qname, err = dns.ReverseAddr(q.Name)
 		// might be an actual DNS name instead of an IP address
 		// if that looks likely, use it as is
-		if err != nil && !util.IsStringValidDomainName(qname) {
-			return nil, nil, StatusIllegalInput, err
-			// q.Name is a valid name, we can continue
+		if err != nil {
+			if !util.IsStringValidDomainName(q.Name) {
```
phillip-stephens: Seems like a good fix, the only thing I don't _love_ is it's a bit hard to reason about the case where `err != nil` but `util.IsStringValidDomainName == true`. In that case, we skip the `else` which might not be best.

I'm thinking this will be flatter logically and thus easier to read
```go
		if err != nil && !util.IsStringValidDomainName(q.Name) {
			return nil, nil, StatusIllegalInput, err
		} else if len(qname) > 0 && qname[len(qname)-1] == '.' {
			// remove trailing "." added by dns.ReverseAddr
			q.Name = qname[:len(qname)-1]
		}
```

If you agree, I'll make the fix. Tested it with the sample input and looks good 👍

zakird: Sure! I don't have a strong opinion on the logic.

phillip-stephens (approved):
LGTM

PR #569, opened 2025-11-03
Add functionality to parse zone files

### LINKED ISSUE #568: Parse zone file input
Over the last year, one of my use cases for `zdns` has required being able to query entire zones. My workaround for this has been a pre-processing script which I had written to extract hostnames from zone files, but I think this use case (however niche) might still be useful enough to justify including a parser within `zdns`. 

**Solution**
The implementation is straightforward.
- `cli/iohandlers/zonefile_handler.go` will have the parsing logic;
- additional logic in `cli/worker_manager.go` (e.g., a zone file can't also be an Alexa file, etc.);
- two new entries to `InputOutputOptions` struct in `cli.go`

**Usage example**
- `zdns A --zone-file -f in.txt`
- an additional flag `--zone-file-include-targets` (bool `false` by default) will be available for extracting additional domains, i.e., those found within `rdata` (for NS, CNAME, DNAME, or PTR record types);

I've already written the code for this and it works fine for my use case. If this is a welcome feature, I'll create a pull request.

qurbat:
In reference to the issue I had created earlier: #568
- **src/cli/iohandlers/zonefile_handler.go** defines the parsing logic. I used the zone file for a less popular TLD (which I retrieved from ICANN's CZDS portal) as a reference. This could be more robust, but I think it should be fine for now.
- **src/cli/worker_manager.go** defines additional logic (e.g., a zone file can't also be an Alexa file, etc.);
- **src/cli/cli.go** the `InputOutputOptions` struct is expanded for `--zone-file` and `--include-targets`
- **testing/integration_tests.py** defines two tests for the parsing functions 

Resolves #568

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -107,6 +107,8 @@ type InputOutputOptions struct {
 	ResultVerbosity              string `long:"result-verbosity" default:"normal" description:"Sets verbosity of each output record. Options: short, normal, long, trace"`
 	StatusUpdatesFilePath        string `short:"u" long:"status-updates-file" default:"-" description:"file to write scan progress to, defaults to stderr"`
 	Verbosity                    int    `long:"verbosity" default:"3" description:"log verbosity: 1 (lowest)--5 (highest)"`
+	ZoneFileFormat               bool   `long:"zone-file" description:"input file is a DNS zone file"`
+	ZoneFileIncludeTargets       bool   `long:"include-targets" description:"include target domains from zone file records (e.g., nameservers from NS records). Only used with --zone-file"`
 }
 
 type CLIConf struct {
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -108,7 +108,7 @@ type InputOutputOptions struct {
 	StatusUpdatesFilePath        string `short:"u" long:"status-updates-file" default:"-" description:"file to write scan progress to, defaults to stderr"`
 	Verbosity                    int    `long:"verbosity" default:"3" description:"log verbosity: 1 (lowest)--5 (highest)"`
 	ZoneFileFormat               bool   `long:"zone-file" description:"input file is a DNS zone file"`
-	ZoneFileIncludeTargets       bool   `long:"include-targets" description:"include target domains from zone file records (e.g., nameservers from NS records). Only used with --zone-file"`
+	ZoneFileIncludeTargets       bool   `long:"zone-file-include-targets" description:"include target domains from zone file records (e.g., nameservers from NS records). Only used with --zone-file"`
 }
 
 type CLIConf struct {
--- a/src/cli/iohandlers/zonefile_handler.go
+++ b/src/cli/iohandlers/zonefile_handler.go
@@ -58,9 +58,9 @@ func (h *ZoneFileInputHandler) parseZoneLine(line string) []string {
 	if name != "" {
 		domains = append(domains, name)
 	}
-if !h.includeTargets {
-    return domains
-}
+	if !h.includeTargets {
+		return domains
+	}
 	// Find the record type by skipping TTL (numeric) and CLASS (IN)
 	// Standard format: NAME TTL CLASS TYPE RDATA
 	typeIndex := -1
--- a/src/cli/iohandlers/zonefile_handler.go
+++ b/src/cli/iohandlers/zonefile_handler.go
@@ -1,0 +1,155 @@
+/*
+ * ZDNS Copyright 2024 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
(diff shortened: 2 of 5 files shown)


qurbat:
Lint is failing, but it's in `status_handler.go` which I haven't changed. 

```
karan@mbp zdns-parse-host-files % make lint
goimports -w -local "github.com/zmap/zdns" ./
gofmt -s -w ./
golangci-lint run
WARN [runner/nolint_filter] Found unknown linters in //nolint directives: golint 
src/cli/iohandlers/status_handler.go:167:3: concat-loop: string concatenation in a loop (perfsprint)
		returnStr += fmt.Sprintf("%s: %d, ", statusOccurrence.status, statusOccurrence.occurrence)
		^
1 issues:
* perfsprint: 1
make: *** [lint] Error 1
```

@zakird do you know why this might be happening?

phillip-stephens:
I can take care of that lint error, likely the linter we use was updated with an additional rule and so it's breaking in `main` as well.

qurbat:
> I can take care of that lint error, likely the linter we use was updated with an additional rule and so it's breaking in `main` as well.

Thanks @phillip-stephens!

@zakird There was another failing lint check for the integration test but I've pushed another commit for that. All checks are passing now.

src/cli/cli.go
```diff
@@ -107,6 +107,8 @@ type InputOutputOptions struct {
 	ResultVerbosity              string `long:"result-verbosity" default:"normal" description:"Sets verbosity of each output record. Options: short, normal, long, trace"`
 	StatusUpdatesFilePath        string `short:"u" long:"status-updates-file" default:"-" description:"file to write scan progress to, defaults to stderr"`
 	Verbosity                    int    `long:"verbosity" default:"3" description:"log verbosity: 1 (lowest)--5 (highest)"`
+	ZoneFileFormat               bool   `long:"zone-file" description:"input file is a DNS zone file"`
+	ZoneFileIncludeTargets       bool   `long:"include-targets" description:"include target domains from zone file records (e.g., nameservers from NS records). Only used with --zone-file"`
```
phillip-stephens: Let's move this line up to alphabetize the flags.

phillip-stephens:
```suggestion
	ZoneFileIncludeTargets       bool   `long:"zone-file-include-targets" description:"include target domains from zone file records (e.g., nameservers from NS records). Only used with --zone-file"`
```

Think this would make the usage of the flag just a little clearer.

src/cli/iohandlers/zonefile_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	domains := make([]string, 0, 2)
+
+	name := strings.ToLower(fields[0])
+	name = strings.TrimSuffix(name, ".")
+	if name != "" {
+		domains = append(domains, name)
+	}
+
```
phillip-stephens:
```suggestion
if !h.includeTargets {
    return domains
}
```
Believe we can return early here and save the other field parsing if we don't need to include targets as well.

src/cli/worker_manager.go
```diff
@@ -143,6 +143,18 @@ func populateCLIConfig(gc *CLIConf) *CLIConf {
 	if gc.NameServerMode && gc.MetadataFormat {
 		log.Fatal("Metadata mode is incompatible with name server mode")
 	}
+	if gc.NameServerMode && gc.ZoneFileFormat {
+		log.Fatal("Zone file mode is incompatible with name server mode")
+	}
+	if gc.ZoneFileIncludeTargets && !gc.ZoneFileFormat {
+		log.Fatal("--include-targets can only be used with --zone-file")
```
phillip-stephens: If you take my earlier flag name change, don't forget to update here too.

phillip-stephens (changes requested):
Really appreciate this PR, I think this will be a really useful feature for people!

Had a few comments and then we should write up a lil blurb in our `Input Formats` README [section](https://github.com/zmap/zdns?tab=readme-ov-file#input-formats). 

Something like the following:
## Zone File
Zone files (from ICANN CZDS or similar) can be used as in input source for ZDNS with the `--zone-file` flag. This enables parsing zone files from either `stdin` (default) or a file with the `--input-file` flag.

By default, ZDNS will extract only the `name` from each zone file record. If you'd also like to resolve the names referenced in the `answer` section of record types such as `CNAME`s or `NS` records, you can use the `--zone-file-include-targets` CLI flag.

For example, with `--zone-file-include-targets` and this DNS zone file entry:
```shell
example.com. 3600 IN NS ns1.example.com
```
both `example.com` and `ns1.example.com` would be resolved.

qurbat:
@phillip-stephens Thanks for the review Phillip! I'll make these changes over the weekend.

qurbat:
@phillip-stephens I've resolved your comments. Please check again once when you get the time. Thanks!

phillip-stephens (approved):
Thanks @qurbat for the great work!

PR #612, opened 2026-06-17
fix README links

rewrite:
Previously, the links in the readme were broken and linked to nonexistent locations (e.g. examples -> `https://github.com/zmap/zdns/blob/main/github.com/zmap/zdns/examples`). This has been fixed in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,13 @@ make install
 Usage
 =====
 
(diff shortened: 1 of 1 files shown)


phillip-stephens (approved):
Good catch and thanks for the contribution!

PR #614, opened 2026-07-02
[svcb.dohpath] add dohpath extraction when running SVCB queries

mkoch-tud:
When querying SVCB records with ZDNS, it is necessary to also extract the DoH path if given in the reply.
That allows the user to get the right endpoint for a DoH query (especially useful when a non standard endpoint is used).

The SVCBDoHPath is already present in the dns module, it was just not used during evaluation.


zakird (approved):

zakird:
Thank you!!

PR #618, opened 2026-07-16
Fall back to non-EDNS queries on FORMERR

### LINKED ISSUE #617: [Bug] Iterative resolver does not retry without EDNS when an authoritative server returns FORMERR
**Describe the bug**

When ZDNS is used as a Go library for iterative resolution, it cannot resolve names served by an authoritative nameserver that does not support EDNS.

ZDNS unconditionally adds an EDNS0 OPT record to its wire queries. If the authoritative server responds with `FORMERR` and does not include an OPT record, ZDNS returns the failure instead of retrying the same query without EDNS.

This causes otherwise valid and resolvable names to fail. For example, an iterative lookup of `6thfamily-com.mail.protection.outlook.com` reaches the authoritative nameservers for `mail.protection.outlook.com`, but the EDNS query receives a 12-byte `FORMERR` response. Repeating the query without EDNS returns an authoritative `NOERROR` response with the expected A records.

This was encountered through the ZDNS Go library, not through the ZDNS CLI.

**To Reproduce**

- ZDNS version: the behavior is present at `main` commit `5edb8b98ef249128d947d9b6bded158b442dbfb6` (`v2.1.1-7-g5edb8b9`)
- Usage: ZDNS Go library (`github.com/zmap/zdns/v2`)
- Lookup mode: iterative
- Input: `6thfamily-com.mail.protection.outlook.com`
- `/etc/resolv.conf`: not applicable; this is an iterative lookup starting from the configured root nameservers

A minimal library lookup is equivalent to:

<pre>
package main

import (
    "context"
    "fmt"
    "log"

    "github.com/zmap/dns"
    "github.com/zmap/zdns/v2/src/zdns"
)

func main() {
    config := zdns.NewResolverConfig()

    resolver, err := zdns.InitResolver(config)
    if err != nil {
        log.Fatal(err)
    }

    result, trace, status, err := resolver.IterativeLookup(
        context.Background(),
        &amp;zdns.Question{
            Name:  "6thfamily-com.mail.protection.outlook.com",
            Type:  dns.TypeA,
            Class: dns.ClassINET,
        },
    )

    fmt.Printf("status=%s err=%v\n", status, err)
    fmt.Printf("result=%+v\n", result)
    fmt.Printf("trace=%+v\n", trace)
}
</pre>

The authoritative-server behavior can also be reproduced independently with `dig`:

<pre>
$ dig 6thfamily-com.mail.protection.outlook.com \
    @ns1-proddns.glbdns.protection.outlook.com. \
    +norecurse

;; -&gt;&gt;HEADER&lt;&lt;- opcode: QUERY, status: FORMERR
;; flags: qr; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: EDNS query returned status FORMERR - retry with '+noedns'
;; MSG SIZE  rcvd: 12
</pre>

The same query without EDNS succeeds:

<pre>
$ dig 6thfamily-com.mail.protection.outlook.com \
    @ns1-proddns.glbdns.protection.outlook.com. \
    +norecurse \
    +noedns

;; -&gt;&gt;HEADER&lt;&lt;- opcode: QUERY, status: NOERROR
;; flags: qr aa; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;6thfamily-com.mail.protection.outlook.com. IN A

;; ANSWER SECTION:
6thfamily-com.mail.protection.outlook.com. 10 IN A 52.101.41.56
6thfamily-com.mail.protection.outlook.com. 10 IN A 52.101.10.10
6thfamily-com.mail.protection.outlook.com. 10 IN A 52.101.194.0
6thfamily-com.mail.protection.outlook.com. 10 IN A 52.101.41.24
</pre>

The exact returned addresses may vary, but the relevant difference is stable: the EDNS query returns `FORMERR`, while the non-EDNS query returns an authoritative answer.

A full `dig +trace` also reaches the Microsoft authoritative tier and then stops after receiving the 12-byte response:

<pre>
$ dig 6thfamily-com.mail.protection.outlook.com +trace

...
mail.protection.outlook.com. 3600 IN NS ns1-proddns.glbdns.protection.outlook.com.
mail.protection.outlook.com. 3600 IN NS ns2-proddns.glbdns.protection.outlook.com.
...
;; Received 12 bytes from 52.102.36.182#53(ns1-proddns.glbdns.protection.outlook.com)
</pre>

**Expected behavior**

For an ordinary lookup that does not require DNSSEC or an explicitly requested EDNS option, ZDNS should detect an EDNS-intolerant response and retry the query against the same nameserver without an OPT record.

In particular, a `FORMERR` response without an OPT record indicates that the remote server may not support EDNS. The expected sequence is:

1. Send the normal EDNS-enabled query.
2. Receive `FORMERR` without an OPT record.
3. Retry the same question against the same nameserver without EDNS.
4. Continue iterative resolution using the successful authoritative response.

ZDNS should not perform this fallback when EDNS is required for the lookup, such as when DNSSEC is enabled. A `FORMERR` response that includes an OPT record should also remain distinguishable from a response indicating that the server does not support EDNS.

**Screenshots**

Not applicable. The textual DNS responses above show the failure.

**Additional context**

All current wire-query paths call `SetEdns0(1232, ...)` unconditionally, including UDP, TCP, DNS-over-TLS, and DNS-over-HTTPS. There does not appear to be a path that constructs a query without an OPT record.

Additionally, `FORMERR` is not included in `isStatusRetryable`, so iterative resolution returns the error without attempting an EDNS downgrade. Other retry paths still construct another EDNS-enabled query, and the UDP-to-TCP fallback also retains EDNS.

RFC 6891 describes fallback to unextended DNS when the remote endpoint does not support EDNS. It also distinguishes:

- `FORMERR` without OPT, which can indicate that the responder does not implement EDNS; and
- `FORMERR` with OPT, which indicates a problem processing the EDNS request itself.

See [RFC 6891 sections 6.2.2 and 7](https://www.rfc-editor.org/rfc/rfc6891.html#section-6.2.2).

lattwood:
## What

ZDNS always adds an EDNS OPT record to queries. Some authoritative servers return `FORMERR` when EDNS is present but answer the same query normally without it.

If we get `FORMERR` without an OPT record in the response, retry the same nameserver without EDNS.

There are a few guardrails:

- Don’t fall back when DNSSEC is enabled.
- Don’t fall back when explicit EDNS options were requested.
- Don’t fall back when the `FORMERR` response contains OPT—the server understood EDNS in that case.
- Give the fallback its own rate-limit admission and network timeout.
- Don’t cache EDNS support, since that can leave long-running resolvers stuck with stale capability state.

This applies to UDP, TCP, DoT, and DoH, and leaves the existing UDP-to-TCP truncation behavior alone.

## Testing

Added coverage for:

- UDP and TCP fallback.
- `FORMERR` responses with and without OPT.
- DNSSEC and explicit EDNS options.
- Rate-limit admission for both attempts.
- A delayed `FORMERR` followed by a delayed successful response through `ExternalLookup`, verifying that both wire operations get a full `NetworkTimeout`.

Also ran:

- `go test ./...`
- Focused tests under the race detector.
- `golangci-lint run ./...`
- The original query against the Microsoft authoritative server.

Fixes #617

### DIFF
--- a/src/zdns/lookup.go
+++ b/src/zdns/lookup.go
@@ -891,25 +891,15 @@ func (r *Resolver) cachedLookup(ctx context.Context, q Question, nameServer *Nam
 	var result *SingleQueryResult
 	var rawResp *dns.Msg
 	var status Status
-	if r.dnsOverHTTPSEnabled {
-		r.verboseLog(depth, "****WIRE LOOKUP*** ", DoHProtocol, " ", dns.TypeToString[q.Type], " ", q.Name, " ", nameServer)
-		result, rawResp, status, err = doDoHLookup(lookupCtx, connInfo.httpsClient, q, nameServer, requestIteration, r.ednsOptions, r.dnsSecEnabled, r.checkingDisabledBit, r.authenticatedDataBit)
-	} else if r.dnsOverTLSEnabled {
-		r.verboseLog(depth, "****WIRE LOOKUP*** ", DoTProtocol, " ", dns.TypeToString[q.Type], " ", q.Name, " ", nameServer)
-		result, rawResp, status, err = doDoTLookup(lookupCtx, connInfo, q, nameServer, r.rootCAs, r.verifyServerCert, requestIteration, r.ednsOptions, r.dnsSecEnabled, r.checkingDisabledBit, r.authenticatedDataBit)
-	} else if connInfo.udpClient != nil {
-		r.verboseLog(depth, "****WIRE LOOKUP*** ", UDPProtocol, " ", dns.TypeToString[q.Type], " ", q.Name, " ", nameServer)
-		result, rawResp, status, err = wireLookupUDP(lookupCtx, connInfo, q, nameServer, r.ednsOptions, requestIteration, r.dnsSecEnabled, r.checkingDisabledBit, r.authenticatedDataBit)
-		if status == StatusTruncated && connInfo.tcpClient != nil {
-			// result truncated, try again with TCP
-			r.verboseLog(depth, "****WIRE LOOKUP*** ", TCPProtocol, " ", dns.TypeToString[q.Type], " ", q.Name, " ", nameServer)
-			result, rawResp, status, err = wireLookupTCP(lookupCtx, connInfo, q, nameServer, r.ednsOptions, requestIteration, r.dnsSecEnabled, r.checkingDisabledBit, r.authenticatedDataBit)
+	result, rawResp, status, err = r.performWireLookup(lookupCtx, connInfo, q, nameServer, requestIteration, depth, true)
+	if err == nil && shouldRetryWithoutEDNS(rawResp, r.ednsOptions, r.dnsSecEnabled) {
+		fallbackCtx, fallbackCancel := context.WithTimeout(ctx, r.networkTimeout)
+		defer fallbackCancel()
+
+		if rateLimitErr := r.rateLimit.wait(fallbackCtx, *nameServer); rateLimitErr != nil {
+			return &SingleQueryResult{}, false, StatusError, trace, fmt.Errorf("rate limiter error for EDNS fallback against nameserver %s: %w", nameServer, rateLimitErr)
 		}
-	} else if connInfo.tcpClient != nil {
-		r.verboseLog(depth, "****WIRE LOOKUP*** ", TCPProtocol, " ", dns.TypeToString[q.Type], " ", q.Name, " ", nameServer)
-		result, rawResp, status, err = wireLookupTCP(lookupCtx, connInfo, q, nameServer, r.ednsOptions, requestIteration, r.dnsSecEnabled, r.checkingDisabledBit, r.authenticatedDataBit)
-	} else {
-		return &SingleQueryResult{}, false, StatusError, trace, errors.New("no connection info for nameserver")
+		result, rawResp, status, err = r.performWireLookup(fallbackCtx, connInfo, q, nameServer, requestIteration, depth, false)
 	}
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


yarrrly:
Not a maintainer. I was looking at how DNS measurement tools handle EDNS-intolerant servers and ended up reading this, so two notes that may save review time.

The condition still reproduces, and it is wider than the one name in #617. Every `*.mail.protection.outlook.com` name I tried answers FORMERR to an EDNS query and answers normally without one, on both authoritative servers (2026-08-20):

```
$ dig +norecurse microsoft-com.mail.protection.outlook.com @ns1-proddns.glbdns.protection.outlook.com
;; ->>HEADER<<- opcode: QUERY, status: FORMERR, id: 56406
;; flags: qr; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

$ dig +norecurse +noedns microsoft-com.mail.protection.outlook.com @ns1-proddns.glbdns.protection.outlook.com
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 21361
;; ANSWER SECTION:
microsoft-com.mail.protection.outlook.com. 10 IN A 52.101.10.2
microsoft-com.mail.protection.outlook.com. 10 IN A 52.101.60.68
microsoft-com.mail.protection.outlook.com. 10 IN A 52.101.50.10
```

Same for `harvard-edu.` and the reporter's `6thfamily-com.`, against ns1 and ns2. `microsoft.com`'s own MX target is one of these names, so this is the Microsoft 365 mail-routing tier rather than a single misconfigured server, and any iterative scan that resolves MX targets meets it.

Two things in the diff I had to convince myself of:

* `shouldRetryWithoutEDNS` is narrower than "retry on FORMERR". It requires no OPT in the response and no DNSSEC and no EDNS options in the request, which is the RFC 6891 section 7 case and nothing beyond it. The FORMERR-with-OPT case has its own table entry asserting that no retry happens.
* `useEDNS` is threaded through the truncation path, so a plain UDP answer that comes back truncated retries over TCP still without an OPT record instead of quietly re-adding EDNS.

The fallback spends a second rate-limiter token against that nameserver, which the tests assert. Only servers that answer FORMERR-without-OPT pay it, which for a scanner seems the right place for the cost to land.

zakird (approved):

zakird:
Thank you!!

